PR #3, opened 2016-10-12
Infrastructure sprint updates (.Net 4.6.1, etc.)

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/3)


hatton:
LGTM Tom. As with all these other projects, I'm concerned that we're going to be publishing releases that change user requirements, without finding/updating the place(s) we list them.

tombogle:
John, when you say "user requirements" are you referring to the minimum hardware/OS requirements/dependencies, or something else?

PR #7, opened 2017-05-31
This is an attempt to see if removing NUnitForms and the tests that r…

tombogle:
…ely on it will result in a successfull build on TeamCity and get past the problem with loading SIL.Core.dll

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/7)


tombogle:
Didn't solve anything

PR #10, opened 2017-06-12
Automatic wix installer

GlennPruitt:


hatton:
Let's gitignore the binaries.

GlennPruitt:
I added lib/dotnet/Palaso*.dll to the gitignore file.

hatton:
Thanks Glenn. These dlls are still committed though. Perhaps you can squash your 4 commits and re-push, and that will remove them.

PR #15, opened 2018-02-15
Reference the community help site here.

gtryus:
This change would allow users to know where to go for support
rather than going to the SIL site and looking for the link there.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/15)


gtryus:
Moved to a different branch

PR #19, opened 2018-02-27
st378 Avoid setting Actor language to undefined.

gtryus:
When the project is exported the language used by the project is
retained and used if the language is undefined.

This change depends on https://github.com/sillsdev/libpalaso/pull/624
Card: trello.com/c/2daw3H2s

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/19)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 obj/*
 output/*
+.vs*
 *.suo
 *.user
 _ReSharper*
@@ -65,7 +66,7 @@ Thumbs.db
 src/AutoSegmenter/obj/x86/Debug/DesignTimeResolveAssemblyReferencesInput.cache
 lib/dotnet/L10NSharp*
 lib/dotnet/SilUtils.dll
-lib/dotnet/SIL.Archiving.*
+lib/dotnet/SIL.*
 lib/dotnet/icu*
 lib/dotnet/Interop*
 lib/dotnet/Ionic*
@@ -82,5 +83,5 @@ packages/Analytics*
 lib/dotnet/SIL.*.dll
 lib/dotnet/SIL.*.pdb
 .nuget/NuGet.exe
-build/SIL.BuildTasks.dll
+build/SIL.BuildTasks.*
 lib/dotnet/NAudio.dll
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -21,6 +21,8 @@ namespace SayMore.Model
 {
 	static class ArchivingHelper
 	{
+		internal static ArchivingLanguage _defaultLanguage;
+
 		/// ------------------------------------------------------------------------------------
 		internal static void ArchiveUsingIMDI(IIMDIArchivable element)
 		{
@@ -194,20 +196,15 @@ namespace SayMore.Model
 				var actor = InitializeActor(model, person, sessionDateTime);
 
 				// do this to get the ISO3 codes for the languages because they are not in saymore
-				var language = LanguageList.FindByEnglishName(person.MetaDataFile.GetStringValue("primaryLanguage", null));
-				if (language != null)
-					actor.PrimaryLanguage = new ArchivingLanguage(language.Iso3Code, language.EnglishName);
-
-				language = LanguageList.FindByEnglishName(person.MetaDataFile.GetStringValue("mothersLanguage", null));
-				if (language != null)
-					actor.MotherTongueLanguage = new ArchivingLanguage(language.Iso3Code, language.EnglishName);
+				actor.PrimaryLanguage = GetOneLanguage(person.MetaDataFile.GetStringValue("primaryLanguage", null));
+				actor.MotherTongueLanguage = GetOneLanguage(person.MetaDataFile.GetStringValue("mothersLanguage", null));
 
 				// otherLanguage0 - otherLanguage3
 				for (var i = 0; i < 4; i++)
 				{
-					language = LanguageList.FindByEnglishName(person.MetaDataFile.GetStringValue("otherLanguage" + i, null));
-					if (language != null)
-						actor.Iso3Languages.Add(new ArchivingLanguage(language.Iso3Code, language.EnglishName));
+					var languageKey = person.MetaDataFile.GetStringValue("otherLanguage" + i, null);
+					if (string.IsNullOrEmpty(languageKey)) continue;
+					actor.Iso3Languages.Add(GetOneLanguage(languageKey));
 				}
 
 				// custom person fields
@@ -255,6 +252,28 @@ namespace SayMore.Model
 			var files = saymoreSession.GetSessionFilesToArchive(model.GetType());
 			foreach (var file in files)
 				imdiSession.AddFile(CreateArchivingFile(file));
+		}
+
+		internal static ArchivingLanguage GetOneLanguage(string languageKey)
+		{
+			ArchivingLanguage returnValue = null;
+			var language = LanguageList.FindByEnglishName(languageKey);
+			if (language == null || language.Iso3Code == "und")
+				if (!string.IsNullOrEmpty(languageKey) && languageKey.Length == 3)
+					language = LanguageList.FindByISO3Code(languageKey);
+			if (language != null && language.Iso3Code != "und" && !string.IsNullOrEmpty(language.EnglishName))
+				returnValue = new ArchivingLanguage(language.Iso3Code, language.Definition)
+				{
+					EnglishName = language.EnglishName
+				};
+			else if (_defaultLanguage != null)
+			{
+				returnValue = new ArchivingLanguage(_defaultLanguage.Iso3Code, _defaultLanguage.LanguageName)
+				{
+					EnglishName = _defaultLanguage.EnglishName
+				};
+			}
+			return returnValue;
 		}
 
 		internal static ArchivingActor InitializeActor(ArchivingDlgViewModel model, Person person, DateTime sessionDateTime)
@@ -353,10 +372,11 @@ namespace SayMore.Model
 					var language = LanguageList.FindByISO3Code(parts[0]);
 
 					// SP-765:  Allow codes from Ethnologue that are not in the Arbil list
-					if ((language == null) || (string.IsNullOrEmpty(language.EnglishName)))
-						package.ContentIso3Languages.Add(new ArchivingLanguage(parts[0], parts[1], parts[1]));
+					if (string.IsNullOrEmpty(language?.EnglishName))
+						_defaultLanguage = new ArchivingLanguage(parts[0], parts[1], parts[1]);
 					else
-						package.ContentIso3Languages.Add(new ArchivingLanguage(language.Iso3Code, parts[1], language.EnglishName));
+						_defaultLanguage = new ArchivingLanguage(language.Iso3Code, parts[1], language.EnglishName);
+					package.ContentIso3Languages.Add(_defaultLanguage);
 				}
 			}
 
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -170,7 +170,7 @@ namespace SayMore.Model
 			// session situation
 			stringVal = saymoreSession.MetaDataFile.GetStringValue("situation", null);
 			if (!string.IsNullOrEmpty(stringVal))
-				imdiSession.AddKeyValuePair("Situation", stringVal);
+				imdiSession.AddContentKeyValuePair("Situation", stringVal);
 
 			imdiSession.Genre = GetFieldValue(sessionFile, "genre");
 			imdiSession.SubGenre = GetFieldValue(sessionFile, "additional_Sub-Genre");
@@ -183,7 +183,7 @@ namespace SayMore.Model
 
 			// custom session fields
 			foreach (var item in saymoreSession.MetaDataFile.GetCustomFields())
-				imdiSession.AddKeyValuePair(item.FieldId, item.ValueAsString);
+				imdiSession.AddContentKeyValuePair(item.FieldId, item.ValueAsString);
 
 			// actors
 			var actors = new ArchivingActorCollection();
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -196,15 +196,19 @@ namespace SayMore.Model
 				var actor = InitializeActor(model, person, sessionDateTime);
 
 				// do this to get the ISO3 codes for the languages because they are not in saymore
-				actor.PrimaryLanguage = GetOneLanguage(person.MetaDataFile.GetStringValue("primaryLanguage", null));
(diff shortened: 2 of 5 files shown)


hatton:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/SayMore/Model/ArchivingHelper.cs, line 200 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6N_dX7rx4YzkuHsPis:-L6N_dX8LufJ4tjmuj5f:b-ehc87g) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMore/Model/ArchivingHelper.cs#L200)):*
> ```Smalltalk
> 				// do this to get the ISO3 codes for the languages because they are not in saymore
> 				actor.PrimaryLanguage = GetOneLanguage(person.MetaDataFile.GetStringValue("primaryLanguage", null));
> 				actor.MotherTongueLanguage = GetOneLanguage(person.MetaDataFile.GetStringValue("mothersLanguage", null));
> ```

This isn't a change here, but I probably didn't review the original. I think it might be wrong. It takes "mothertongue" to literally mean "the tongue spoken by the mother".  As intuitive as that sounds, I think it is not correct. By my defnition, this would be "Native Speaker language", which  is what is the language the person grew up speaking, which in many cases may not include much of the mother's language. There are other definitions: I just learned that in many cases, it may actually mean "the language of my ethinic group", which may or may not be something I speak. https://en.wikipedia.org/wiki/First_language.

So this needs more analysis, and you can ignore my comment for this PR, just wanted to point it out.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 230 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6NZEHVHBDjBS8IgAsu:-L6NZEHVHBDjBS8IgAsv:baezd6x) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L230)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void GetOneLanguage_DefinedIsoTest()
> ```

Perhaps there is already a test for this, but it's unusual to use a 3 letter code where a 2 letter one exists. In fact the standard is to use the smaller one. So is this test specifically making sure that if for some reason something went wrong and we had a 3 letter code, things would be ok? Does it make sense to also test for the expected (2 letter) case?

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 248 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6NYi8iyLxCOGey21Kn:-L6NYi8iyLxCOGey21Ko:bdbfji9) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L248)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void GetOneLanguage_UndefinedIsoTest()
> ```

It would be great if you could use the Method_Condition_Expected result pattern for test names, as used elsewhere in this file. I'm looking at these tests and not knowing what it is that they do. For example, if 'tru' is meant to be a code that isn't in the iso3 datbase, how does the program manage to tell me that it is "Turoyo"?

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L6Nbl-gTHU-gyLwBJku:b-tit26s) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*
> ```Smalltalk
> 		}
> 		#endregion
> 	}
> ```

If there isn't one already, could you add a test for an un-registered language (one starting with code 'q') to make sure that works ok for the name lookups? Thx.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/saymore/19)*

gtryus:
> I'm looking at these tests and not knowing what it is that they do. For example, if 'tru' is meant to be a code that isn't in the iso3 datbase, how does the program manage to tell me that it is "Turoyo"?

One common case for languages is that the language being studied is a minority language and the other languages spoken are present in the Max Planck list. So the primary and/or mother tongue language might be the language of the project. At the project level the user has defined the ISO code and the name of this language. The Max Planck list often doesn't have these languages. So in addition to the languages defined on the Max Planck list, this code will assume the language being studied is a defined language.

The test cases also test for the case where the language being studied might not be on the Max Planck list or the language being studied.

If you have a suggestion about how to put this much information into a method name, I can gladly put it in. I wasn't creative enough to identify it.

gtryus:
> So this needs more analysis, and you can ignore my comment for this PR, just wanted to point it out.

Yes this discussion has come up here too but the problem needs to be discussed with Max Planck / SOAS since it is built into their list. They include English as ISO3 eng and French as ISO3 fra and have no reference in the IMDI standard for the two letter codes.

hatton:
Greg, it looks like you're responding to Reviewable comments from github, which makes it hard for the reviewer to go through them. Could go into Reviewable and answer them there? THanks.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/saymore/19#-:-L6PBEsBDJMzJhscdR28:btm4vs5)*

gtryus:
I put some more comments in using renewable and I also updated the code so that the undefined tests are clearer and work consistently no matter what order tests are executed.

---

Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 230 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6NZEHVHBDjBS8IgAsu:-L7RDw_M-kfitZ-f54fP:bq97204) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L230)):*

Perhaps there is already a test for this, but it's unusual to use a 3 letter code where a 2 letter one exists. In fact the standard is to use the smaller one. So is this test specifically making sure that if for some reason something went wrong and we had a 3 letter code, things would be ok? Does it make sense to also test for the expected (2 letter) case?
</blockquote></details>

I think the IMDI standard and Saymore work strictly with 3 letter code. You would need to make some design changes to support two letter codes. We could probably convert three letter codes to two letter codes using SIL.WritingSystems but I don't know if this would be acceptable to Max Planke or Soas.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 248 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6NYi8iyLxCOGey21Kn:-L7REMvJW7ugjJrHf_Q3:bfjzx5b) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L248)):*

It would be great if you could use the Method_Condition_Expected result pattern for test names, as used elsewhere in this file. I'm looking at these tests and not knowing what it is that they do. For example, if 'tru' is meant to be a code that isn't in the iso3 datbase, how does the program manage to tell me that it is "Turoyo"?
</blockquote></details>

Yes, for a three line test it seems the expected result is clear from the code.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L7REWd0awT0uudFQgdr:b9rv6ln) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*

If there isn't one already, could you add a test for an un-registered language (one starting with code 'q') to make sure that works ok for the name lookups? Thx.
</blockquote></details>

I used tru and Turoyo because it was in the Soas example.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/saymore/19#-:-L7RElSMdvI1ioPkeozu:b-eni478)*

gtryus:
Review status: 0 of 2 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/SayMore/Model/ArchivingHelper.cs, line 200 at r1](https://reviewable.io:443/reviews/sillsdev/saymore/19#-L6N_dX7rx4YzkuHsPis:-L7REzwZI-7yp6uxSW8v:b-whg9m) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMore/Model/ArchivingHelper.cs#L200)):*

This isn't a change here, but I probably didn't review the original. I think it might be wrong. It takes "mothertongue" to literally mean "the tongue spoken by the mother".  As intuitive as that sounds, I think it is not correct. By my defnition, this would be "Native Speaker language", which  is what is the language the person grew up speaking, which in many cases may not include much of the mother's language. There are other definitions: I just learned that in many cases, it may actually mean "the language of my ethinic group", which may or may not be something I speak. https://en.wikipedia.org/wiki/First_language.

So this needs more analysis, and you can ignore my comment for this PR, just wanted to point it out.
</blockquote></details>

Yes, it is not clear that this will be simple using the Saymore ui or the differences between it and the IMDI standard. I did note that in their example, more than one language can be marked as "Primary" language for a specific individual. I don't know if this was intentional. In fact they seemed to be "fudging" with the standard a bit in the way the defined the "content" language and "working" language for a particular session.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/saymore/19)*

gtryus:
Review status: 0 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 248 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6NYi8iyLxCOGey21Kn:-L8mfyPP4AAYqjaA2j3G:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L248)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

Yes, for a three line test it seems the expected result is clear from the code.
</blockquote></details>

Done.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L8mg-IS3S-DZB91i9UB:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

I used tru and Turoyo because it was in the Soas example.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/saymore/19)*

hatton:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L8r5JV61MwQcgRaHjwt:bvfxpc6) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*
> q
This is marked as "Done" but I don't see a test that checks for unregistered languages (one starting with code 'q'). Perhaps a mixup with Reviewable?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/saymore/19)*

gtryus:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L9AmGz6BGdjstYEEB50:b-94yrr7) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*

> q
This is marked as "Done" but I don't see a test that checks for unregistered languages (one starting with code 'q'). Perhaps a mixup with Reviewable?
</blockquote></details>

Yes, it must be Reviewable. When you click on "line 276" in the right margin of this interchange, it shows the test "GetOneLanguage_PrivateUseIso_ReturnsNull()". There are also two tests which follow it that are related. Also if from github, you click on the sha abbreviation for the last change set, you will see the latest tests.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/saymore/19)*

hatton:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 230 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6NZEHVHBDjBS8IgAsu:-L9BD6J27bwYu10IqK-a:b-bagl15) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L230)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

I think the IMDI standard and Saymore work strictly with 3 letter code. You would need to make some design changes to support two letter codes. We could probably convert three letter codes to two letter codes using SIL.WritingSystems but I don't know if this would be acceptable to Max Planke or Soas.
</blockquote></details>

Yah, ok... let's not rock the boat.

---

*[src/SayMoreTests/model/SessionArchivingTests.cs, line 276 at r1](https://beta.reviewable.io/reviews/sillsdev/saymore/19#-L6Nbl-gTHU-gyLwBJkt:-L9BD0QMEDas6-16w7BE:b75kui7) ([raw file](https://github.com/sillsdev/saymore/blob/278c5ccb8e770cee515ec2238bf87e9e9a0853d3/src/SayMoreTests/model/SessionArchivingTests.cs#L276)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

Yes, it must be Reviewable. When you click on "line 276" in the right margin of this interchange, it shows the test "GetOneLanguage_PrivateUseIso_ReturnsNull()". There are also two tests which follow it that are related. Also if from github, you click on the sha abbreviation for the last change set, you will see the latest tests.
</blockquote></details>

Thanks Greg.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/saymore/19)*

PR #22, opened 2018-03-14
St384 - output recording device and microphone as recordingEquipment

gtryus:
Get the file object and attach RecordingEquipment keywords for
recording device and microphone.

Depends on: sillsdev/libpalaso#653
card: https://trello.com/c/IlOdLSNy

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/22)

### DIFF
--- a/src/SayMore/Model/Files/ComponentFile.cs
+++ b/src/SayMore/Model/Files/ComponentFile.cs
@@ -135,9 +135,11 @@ namespace SayMore.Model.Files
 		}
 
 		[Obsolete("For Mocking Only")]
-		public ComponentFile()
+		public ComponentFile(FileType fileType)
 		{
 			_componentRoles = ApplicationContainer.ComponentRoles;
+			FileType = fileType;
+			MetaDataFieldValues = new List<FieldInstance>();
 		}
 
 		public void LoadNow()
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -255,7 +255,22 @@ namespace SayMore.Model
 			// session files
 			var files = saymoreSession.GetSessionFilesToArchive(model.GetType());
 			foreach (var file in files)
+			{
 				imdiSession.AddFile(CreateArchivingFile(file));
+				var info = saymoreSession.GetComponentFiles().FirstOrDefault(componentFile => componentFile.PathToAnnotatedFile == file);
+				if (info == null || !info.FileType.IsAudioOrVideo)
+					continue;
+				var device = (from infoValue in info.MetaDataFieldValues
+					where infoValue.FieldId == "Device"
+					select infoValue.ValueAsString).FirstOrDefault();
+				if (!string.IsNullOrEmpty(device))
+					imdiSession.AddFileKeyValuePair(file, "RecordingEquipment", device);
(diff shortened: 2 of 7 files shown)


gtryus:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/SayMore/Model/ArchivingHelper.cs, line 266 at r1](https://reviewable.io/reviews/sillsdev/saymore/22#-L7aQp7IuMYgQdXUQfuV-r1-266:-LBXYA0v6TMeHsKWQWam:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/7d056943352bc1647a3fcfcfc7ae5220219a107b/src/SayMore/Model/ArchivingHelper.cs#L266)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

put the continue on the next line
</blockquote></details>

Done.

---

*[src/SayMore/Model/ArchivingHelper.cs, line 276 at r1](https://reviewable.io/reviews/sillsdev/saymore/22#-L7aQp7IuMYgQdXUQfuV-r1-276:-LBXY0sODJ-BHSH3tlXT:bk81fdf) ([raw file](https://github.com/sillsdev/saymore/blob/7d056943352bc1647a3fcfcfc7ae5220219a107b/src/SayMore/Model/ArchivingHelper.cs#L276)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

This feels odd to add two values with the same file and key string. How does the imdi session handle that?
Also this seems very unit testable and I don't see one.
</blockquote></details>

It works fine in IMDI, It is actually what they have requested and provided in their sample.

I added the unit test.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/22)*

src/SayMore/Model/Files/ComponentFile.cs
```diff
@@ -140,6 +140,14 @@ public ComponentFile()
 			_componentRoles = ApplicationContainer.ComponentRoles;
 		}
 
+		[Obsolete("For Mocking Only")]
```
jasonleenaylor: I'm not sure this change is necessary. The mocking doesn't seem to use it.

gtryus:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/SayMore/Model/Files/ComponentFile.cs, line 143 at r3](https://reviewable.io/reviews/sillsdev/saymore/22#-LBXS38dVw5MyabngPGm-r3-143:-LBXoNRD0ACStKs97PBv:bm62k79) ([raw file](https://github.com/sillsdev/saymore/blob/d99bd7f1f44dec579aa14251bc2f94b85d919f93/src/SayMore/Model/Files/ComponentFile.cs#L143)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

I'm not sure this change is necessary. The mocking doesn't seem to use it.
</blockquote></details>

Yes, the FieldType is passed in the argument in line 333 and the MetaDataFieldValues needs to be initialized for this test.

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/22)*

PR #24, opened 2018-03-15
St385- IMDI .meta files are not part of IMDI export

gtryus:
The information in these files needs to be put into the .imdi file
if it is to be included in the export.

card: https://trello.com/c/A4Z2PJH7

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/24)

### DIFF
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -256,6 +256,8 @@ namespace SayMore.Model
 			var files = saymoreSession.GetSessionFilesToArchive(model.GetType());
 			foreach (var file in files)
 			{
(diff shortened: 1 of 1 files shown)


src/SayMore/Model/ArchivingHelper.cs
```diff
@@ -261,6 +261,7 @@ private static void AddIMDISession(Session saymoreSession, ArchivingDlgViewModel
 			var files = saymoreSession.GetSessionFilesToArchive(model.GetType());
 			foreach (var file in files)
 			{
+				if (file.EndsWith(Settings.Default.MetadataFileExtension, StringComparison.InvariantCulture)) continue;
```
jasonleenaylor: I can see you are matching the style already present in this file but it continues on the same line are ugly. I wouldn't hold this up, but if you highlighted this loop and did code cleanup it would be interesting to see the results.

PR #28, opened 2018-03-20
St391 - Add Project fields to session and resources

gtryus:
Adds Author, Owner, Writes holder, description availability and addresses

Card: https://trello.com/c/aW8AwRH9
Depends on: sillsdev/libpalaso#662

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/28)

### DIFF
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -21,6 +21,7 @@ namespace SayMore.Model
 {
 	static class ArchivingHelper
 	{
+		internal static IMDIPackage _Package;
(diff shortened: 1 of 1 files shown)


src/SayMore/Model/ArchivingHelper.cs
```diff
@@ -21,6 +21,7 @@ namespace SayMore.Model
 {
 	static class ArchivingHelper
 	{
+		internal static IMDIPackage _Package;
```
jasonleenaylor: Couldn't this be done without adding a static member? Either passing in the _Package to methods that would add it or adding it in the calling code?

PR #32, opened 2018-03-29
netCore - update project for latest SIL Core library

gtryus:
SIL Core is using .net core and reporting moved to SIL Desktop
We need to include SIL Desktop with in our build
Interface changes related to st380 are made on SIL Core
see card: https://trello.com/c/IlOdLSNy

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/32)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 obj/*
 output/*
+.vs*
 *.suo
 *.user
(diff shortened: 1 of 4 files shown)


hatton:
:lgtm: but has build failure

PR #36, opened 2018-05-15
Fix: User can specify both content and working lang

manjula-t-ecgroup:
* Added the working language
 * Modified the screen layout

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/36)


gtryus:
Some additional changes applied so this will include the Content and Working languages in the IMDI package. See: https://github.com/sillsdev/saymore/pull/38

PR #41, opened 2018-05-22
SP-1815 Crash deleting custom field

vkarthim:
*Added logic to avoid inserting null value into the model
*Fixed on this scenario based

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/41)

### DIFF
--- a/src/SayMore/UI/ComponentEditors/FieldsValuesGrid.cs
+++ b/src/SayMore/UI/ComponentEditors/FieldsValuesGrid.cs
@@ -344,8 +344,21 @@ namespace SayMore.UI.ComponentEditors
 			else
 			{
 				string oldId;
+
 				if (!_model.GetShouldAskToRemoveFieldEverywhere(e.RowIndex, e.Value as string, out oldId))
-					_model.SaveIdForIndex(e.RowIndex, e.Value as string);
+				{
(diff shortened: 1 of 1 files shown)


src/SayMore/UI/ComponentEditors/FieldsValuesGrid.cs
```diff
@@ -344,8 +344,16 @@ protected override void OnCellValuePushed(DataGridViewCellValueEventArgs e)
 			else
 			{
 				string oldId;
+
 				if (!_model.GetShouldAskToRemoveFieldEverywhere(e.RowIndex, e.Value as string, out oldId))
+				{
+					if (e.Value == null)
```
jasonleenaylor: I think a comment above this null check is needed. 
Explain why we decrementing the RowCount and returning is needed, and why the value might be null.

vkarthim: Ok

gtryus:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/41)*

PR #46, opened 2018-06-05
Session updated so access level information can be retained for each …

ghost:
…archive

*Migrate AccessProtocol code added at Project level
*Added event for Access ComboBox
*Added code to handle AccessProtocol insert and update
*Added code to handle AccessProtocol change from ComboBox
*Added Test case for this issue

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/46)

### DIFF
--- a/src/SayMore/ProjectContext.cs
+++ b/src/SayMore/ProjectContext.cs
@@ -46,7 +46,7 @@ namespace SayMore
 
 			Project = _scope.Resolve<Func<string, Project>>()(projectSettingsPath);
 
-			SetContributorsListToSession(Project.SessionsFolder);
+			SetContributorsListAndMigrateAccessProtocolToSession(Project.SessionsFolder, Project.AccessProtocol);
 
 			var peopleRepoFactory = _scope.Resolve<ElementRepository<Person>.Factory>();
 			peopleRepoFactory(rootDirectoryPath, Person.kFolderName, _scope.Resolve<PersonFileType>());
@@ -80,13 +80,14 @@ namespace SayMore
 			ProjectWindow = _scope.Resolve<ProjectWindow.Factory>()(projectSettingsPath, views);
 		}
 
-		///-------------------------------------------------------------------------------------------------------
-		/// <summary>
-		/// Set the contributor list to the session file from the metafiles
-		/// </summary>
-		/// <param name="sessionsFolder">Session folder path</param>
-		///-------------------------------------------------------------------------------------------------------
-		public static void SetContributorsListToSession(string sessionsFolder)
+		/// -------------------------------------------------------------------------------------------------------
+		///  <summary>
+		///  Set the contributor list to the session file from the metafiles
+		///  </summary>
+		///  <param name="sessionsFolder">Session folder path</param>
+		/// <param name="accessProtocol">Current AccessProtocol Name</param>
+		/// -------------------------------------------------------------------------------------------------------
+		public static void SetContributorsListAndMigrateAccessProtocolToSession(string sessionsFolder, string accessProtocol)
 		{
 			if (!Directory.Exists(sessionsFolder) || Path.GetFileName(sessionsFolder).ToLower() != "sessions")
 			{
@@ -134,6 +135,23 @@ namespace SayMore
 				newContributionsNode.SetAttribute("type", "xml");
 				newContributionsNode.InnerXml = contributorLists.ToString();
 				root.InsertAfter(newContributionsNode, root.LastChild);
+
+				//Insert the AccessProtocol Node with the Access Node
+				var accessProtocolsNode = root?.SelectSingleNode("AccessProtocols") as XmlElement;
+				if (accessProtocolsNode == null)
+				{
+					XmlNode access = root?.SelectSingleNode("access") as XmlElement;
+					accessProtocolsNode = sessionDoc.CreateElement("AccessProtocols");
+					var accessProtocolNode = sessionDoc.CreateElement("AccessProtocol");
+					accessProtocolNode.SetAttribute("Name", accessProtocol);
+					XmlElement accessNode = sessionDoc.CreateElement("access");
+					accessNode.SetAttribute("type", "string");
+					if (access != null) accessNode.InnerText = access.InnerText;
+					accessProtocolNode.AppendChild(accessNode);
+					accessProtocolsNode.AppendChild(accessProtocolNode);
+					root.InsertAfter(accessProtocolsNode, root.LastChild);
+				}
+
 				using (var sessionOutput = XmlWriter.Create(sessionFile, new XmlWriterSettings{Indent = true}))
 				{
 					sessionDoc.Save(sessionOutput);
--- a/src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
+++ b/src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Windows.Forms;
+using System.Xml;
 using L10NSharp;
 using SIL.Extensions;
 using SIL.Reporting;
@@ -17,6 +18,7 @@ using SayMore.Model.Files.DataGathering;
 using SayMore.UI.LowLevelControls;
 using SIL.Archiving.Generic.AccessProtocol;
 using SIL.Archiving.IMDI.Lists;
+using SIL.Windows.Forms.ClearShare;
 
 namespace SayMore.UI.ComponentEditors
 {
@@ -34,6 +36,7 @@ namespace SayMore.UI.ComponentEditors
 		private bool _genreFieldEntered;
 		private List<AccessOption> _accessOptions;
 		private DataGridViewComboBoxEditingControl _moreFieldsComboBox;
+		private bool _isAccessOptionChanged = true;
 
 		/// ------------------------------------------------------------------------------------
 		public SessionBasicEditor(ComponentFile file, string imageKey,
@@ -391,7 +394,6 @@ namespace SayMore.UI.ComponentEditors
 				SetAccessProtocol();
 
 			SetAccessCodeListAndValue();
-
 			base.Activated();
 		}
 
@@ -522,6 +524,8 @@ namespace SayMore.UI.ComponentEditors
 		/// the current value of "access" is no longer in the list</summary>
 		private void SetAccessCodeListAndValue()
 		{
+			_isAccessOptionChanged = false; //To avoid of calling comboBox event
+
 			var currentAccessCode = _file.GetStringValue("access", string.Empty);
 
 			if (_access.DropDownStyle == ComboBoxStyle.DropDown)
@@ -533,22 +537,25 @@ namespace SayMore.UI.ComponentEditors
 			// get the saved list. use .ToList() to copy the list rather than modify the original
 			var choices = _accessOptions.ToList();
 
+			_access.DataSource = choices;
+			_access.DisplayMember = "DisplayMember";
+			_access.ValueMember = "ValueMember";
+
+			currentAccessCode = HandleAccessProtocols(currentAccessCode);
+
 			// is the selected item in the list
 			var found = choices.Any(i => i.ToString() == currentAccessCode);
 
 			if (!found)
 			{
-				choices.Insert(0, new AccessOption { OptionName = "-----" });
-				choices.Insert(0, new AccessOption { OptionName = currentAccessCode });
+				currentAccessCode = _access.Items[0].ToString();
 			}
-
-			_access.DataSource = choices;
-			_access.DisplayMember = "DisplayMember";
-			_access.ValueMember = "ValueMember";
 
 			// select the current code
 			foreach (var item in _access.Items.Cast<object>().Where(item => ((AccessOption)item).ValueMember == currentAccessCode))
 				_access.SelectedItem = item;
+
+			_isAccessOptionChanged = true; //To enable of calling comboBox event
 		}
 
 		/// ------------------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -522,6 +524,8 @@ void backgroundWorker_DoWork(object sender, DoWorkEventArgs e)
 		/// the current value of "access" is no longer in the list</summary>
 		private void SetAccessCodeListAndValue()
 		{
+			_isAccessOptionChanged = false; //To avoid of calling comboBox event
```
tombogle: Unnecessary "of" in comment

papeh:
Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @tombogle and @vkarthim)

---

*[src/SayMore/ProjectContext.cs, line 140 at r1](https://reviewable.io/reviews/sillsdev/saymore/46#-LFcAY8XD4JaDZ9Ocbzq:-LFcAY8XD4JaDZ9Ocbzr:b5enrp) ([raw file](https://github.com/sillsdev/saymore/blob/38b607d33fb3ecdb1c39a91383c846276978a995/src/SayMore/ProjectContext.cs#L140)):*
> ```Smalltalk
> 
> 				//Insert the AccessProtocol Node with the Access Node
> 				var accessProtocolsNode = root?.SelectSingleNode("AccessProtocols") as XmlElement;
> ```

shouldn't need `?.` for  `root.SelectSingleNode`  since we are not using `?.` for `root.InsertAfter`

---

*[src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs, line 527 at r1](https://reviewable.io/reviews/sillsdev/saymore/46#-LFcCbKT90RteihLg9xj:-LFcCbKT90RteihLg9xk:bepwiqr) ([raw file](https://github.com/sillsdev/saymore/blob/38b607d33fb3ecdb1c39a91383c846276978a995/src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs#L527)):*
> ```Smalltalk
> 		private void SetAccessCodeListAndValue()
> 		{
> 			_isAccessOptionChanged = false; //To avoid of calling comboBox event
> ```

instead of creating this new variable, could we deregister and reregister the events we don't want to fire?

---


*Comments from [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/46)*

src/SayMore/ProjectContext.cs
```diff
@@ -46,7 +46,7 @@ public ProjectContext(string projectSettingsPath, IContainer parentContainer)
 
 			Project = _scope.Resolve<Func<string, Project>>()(projectSettingsPath);
 
-			SetContributorsListToSession(Project.SessionsFolder);
+			SetContributorsListAndMigrateAccessProtocolToSession(Project.SessionsFolder, Project.AccessProtocol);
```
tombogle: I realize that this is just a rename, not a new call (as it was apparently added a few weeks ago), but I'm not very sure I would expect a "ProjectContext" constructor to call a method that alters the saved data for another object. Was this design thought out, or was this method call just slapped in here for convenience because it happened to be called at the "right time"?

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+		/// <summary>
+		/// Method to save the new AccessProtocol node and update the existing AccessProtocol node when option get changed
+		/// </summary>
+		/// <param name="accessCode">Previous AccessCode</param>
+		/// <returns>New AccessCode</returns>
+		/// ------------------------------------------------------------------------------------
+		private string HandleAccessProtocols(string accessCode)
+		{
+			string currentAccessCode = accessCode;
+			if (Program.CurrentProject == null) return currentAccessCode;
```
tombogle: I see that SayMore already has quite a few places that don't follow this, but in most of our SIL projects, we generally put the return on the following line, rather than on the same line as the if. I think that generally makes and early return more obvious, resulting in code that is easier to read.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
 		}
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Method to save the new AccessProtocol node and update the existing AccessProtocol node when option get changed
+		/// </summary>
+		/// <param name="accessCode">Previous AccessCode</param>
+		/// <returns>New AccessCode</returns>
+		/// ------------------------------------------------------------------------------------
+		private string HandleAccessProtocols(string accessCode)
```
tombogle: The word "Handle" as the first word in a method name normally signals that the method is some kind of event handler. I think it's best to reserve it for that purpose. Neither this method name nor the description help me understand quickly what it's purpose is.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+		{
+			string currentAccessCode = accessCode;
+			if (Program.CurrentProject == null) return currentAccessCode;
+
+			string sessionFile = _file.PathToAnnotatedFile;
+			var accessProtocol = Program.CurrentProject.AccessProtocol;
+
+			var choices = _accessOptions.ToList();
```
tombogle: _accessOptions is already a list. ToList makes a copy of it (with a name that is shorter but more vague). This method does not appear to be making any changes to the list, so there should be no need to make a copy. I'm guessing this line just got copied from SetAccessCodeListAndValue(), which _does_ modify the list, and therefore needs a copy of it.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+		private void _access_SelectedIndexChanged(object sender, EventArgs e)
+		{
+			if (!_isAccessOptionChanged || Program.CurrentProject == null) return;
+
+			var accessInMetaFile = _file.MetaDataFieldValues.Find(v => v.FieldId == "access");
+			if (accessInMetaFile == null || accessInMetaFile.Value.ToString() == _access.Text)
+				return;
+
+			string sessionFile = _file.PathToAnnotatedFile;
+			var choices = _accessOptions.ToList();
```
tombogle: See comment in previous method regarding ToList.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+			var sessionDoc = new XmlDocument();
+			using (var sessionReader = XmlReader.Create(sessionFile))
+			{
+				sessionDoc.Load(sessionReader);
+			}
+			XmlNode root = sessionDoc.DocumentElement;
+			if (root != null)
+			{
+				XmlNode accessRootNode = root.SelectSingleNode("access");
+				var isAccessProtocolExists = sessionDoc.SelectSingleNode("//AccessProtocols/AccessProtocol[@Name='" + Program.CurrentProject.AccessProtocol + "']");
```
tombogle: Both the prefix "is" and the suffix "Exists" (which should not be used together) suggest that this is a bool, but it's not. I'm guessing you just want to have a variable called "accessProtocol".

src/SayMoreTests/model/ProjectTests.cs
```diff
@@ -347,6 +347,20 @@ public void SetContributorsListToSession_ContributorsInMetaAndSession_AllContrib
...
+		public void SetContributorsListAndMigrateAccessProtocolToSession_MigrateAccessProtocol_NewAccessProtocolNodeCreated()
+		{
+			WriteXmlResource("Qustan059.session");
+			ProjectContext.SetContributorsListAndMigrateAccessProtocolToSession(Path.GetDirectoryName(SessionScenarioPath), "REAP");
+			var sessionXml = new XmlDocument();
+			using (var xmlReader = XmlReader.Create(Path.Combine(SessionScenarioPath, "Qustan059.session")))
+			{
+				sessionXml.Load(xmlReader);
+			}
+			Assert.AreEqual("S", sessionXml.SelectSingleNode("//AccessProtocols").InnerText, "Value should contain inside the AccessProtocols node");
```
tombogle: The comment here should probably explain why we expect an "S" - nothing in the test name or setup suggests what the newly created access protocol node should contain. Does "S" stand for "REAP"?

src/SayMoreTests/model/ProjectTests.cs
```diff
@@ -409,9 +423,9 @@ private void WriteXmlResource(string file)
 				if (!Directory.Exists(SessionScenarioPath))
 					Directory.CreateDirectory(SessionScenarioPath);
 				SessionScenarioPath = Path.Combine(SessionScenarioPath, "ContributorsScenario");
-				if (!Directory.Exists(SessionScenarioPath))
-					Directory.CreateDirectory(SessionScenarioPath);
 			}
+			if (!Directory.Exists(SessionScenarioPath))
```
tombogle: Checking for existence (while a common practice) is technically unnecessary since CreateDirectory will do it again.

src/SayMoreTests/model/ProjectTests.cs
```diff
@@ -347,6 +347,20 @@ public void SetContributorsListToSession_ContributorsInMetaAndSession_AllContrib
 			Assert.AreEqual(4, sessionXml.SelectSingleNode("//participants").InnerText.Split(';').Length, "Expecting four participants (with roles): Iskender Demirel, Lahdo Agirman, Greg Trihus, Eliyo Acar");
 		}
 
+		[Test]
+		[RequiresSTA]
+		public void SetContributorsListAndMigrateAccessProtocolToSession_MigrateAccessProtocol_NewAccessProtocolNodeCreated()
```
tombogle: The fact that this test really only tests the "MigrateAccessProtocol" portion of the method is further evidence to suggest that it should be two different methods, rather than one.

src/SayMoreTests/model/ProjectTests.cs
```diff
@@ -347,6 +347,20 @@ public void SetContributorsListToSession_ContributorsInMetaAndSession_AllContrib
...
+		{
+			WriteXmlResource("Qustan059.session");
+			ProjectContext.SetContributorsListAndMigrateAccessProtocolToSession(Path.GetDirectoryName(SessionScenarioPath), "REAP");
+			var sessionXml = new XmlDocument();
+			using (var xmlReader = XmlReader.Create(Path.Combine(SessionScenarioPath, "Qustan059.session")))
+			{
+				sessionXml.Load(xmlReader);
+			}
+			Assert.AreEqual("S", sessionXml.SelectSingleNode("//AccessProtocols").InnerText, "Value should contain inside the AccessProtocols node");
+		}
```
tombogle: The test should probably also verify the other things this method does, such as setting the "type" attribute to "string".

src/SayMoreTests/model/ProjectTests.cs
tombogle: Looks like this test is for the case where the access protocol does not exist. Another test is needed to test the code path where the access protocol is changed.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+				if (isAccessProtocolExists != null)
+				{
+					string optionSelected = string.Empty;
+					var accessNode = isAccessProtocolExists.SelectSingleNode("access");
+					if (accessNode != null)
+					{
+						optionSelected = choices.FirstOrDefault(i => i.Description.ToString() == _access.Text)?.OptionName;
+						if (optionSelected != null) accessNode.InnerText = optionSelected;
```
tombogle: Again, code will be more readable if you add a line break after the conditional test.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+			{
+				XmlNode accessRootNode = root.SelectSingleNode("access");
+				var isAccessProtocolExists = sessionDoc.SelectSingleNode("//AccessProtocols/AccessProtocol[@Name='" + Program.CurrentProject.AccessProtocol + "']");
+				if (isAccessProtocolExists != null)
+				{
+					string optionSelected = string.Empty;
+					var accessNode = isAccessProtocolExists.SelectSingleNode("access");
+					if (accessNode != null)
+					{
+						optionSelected = choices.FirstOrDefault(i => i.Description.ToString() == _access.Text)?.OptionName;
```
tombogle: Under what circumstances can the user select an access protocol that does not correspond to one of the options?

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -619,5 +626,123 @@ private void HandlePeopleEditClick(object sender, EventArgs e)
...
+
+			var sessionDoc = new XmlDocument();
+			using (var sessionReader = XmlReader.Create(sessionFile))
+			{
+				sessionDoc.Load(sessionReader);
+			}
+			XmlNode root = sessionDoc.DocumentElement;
+			XmlNode accessRootNode = root.SelectSingleNode("access");
+			XmlNode accessProtocolsNode = root.SelectSingleNode("AccessProtocols") ?? sessionDoc.CreateElement("AccessProtocols");
+			var isAccessProtocolExists = root.SelectSingleNode("//AccessProtocols/AccessProtocol[@Name='" + accessProtocol + "']");
```
tombogle: See comment below regarding isAccessProtocolExists variable.

PR #62, opened 2018-08-24
Use DotNetZip. Fix SP-1818. Make build able to restore SilBuildTasks.

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/62)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,14 @@
-obj/*
-output/*
+obj/
+output/**
+**/bin/**
 .vs*
 *.suo
 *.user
 _ReSharper*
-src/Sponge2/bin/*
 design/Mockups 17Nov09.pdf
 del.txt
 bugs.txt
-src/Sponge/bin/*
 src/Sponge2/Sponge2.csproj.John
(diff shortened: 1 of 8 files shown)


tombogle:
___
*[build/SayMore.proj, line 38 at r1](https://reviewable.io/reviews/sillsdev/saymore/62#-LKw2Oc4DnWmOiGs0rYd:-LLaFBKS8W102qJcmxGM:b-13yxit) ([raw file](https://github.com/sillsdev/saymore/blob/d3c54e11b79dccb47067c1ae6ca3ed61dc43e7b2/build/SayMore.proj#L38)):*
<details><summary><i>Previously, gtryus wrote…</i></summary><blockquote>

I am curious about these three lines of messages. Did you intend to remove them before the commit?
</blockquote></details>

Not really. I added the first one specifically to try to debug a problem. One or both of the other ones are included in some of our other projects' build files. They are, of course, just diagnostic messages and could be removed, but they might help with future investigations if something goes awry. I don't think they hurt anything.

PR #73, opened 2018-10-03
SP-1826: Updated/added required packages for compatibility with libpalaso 5

tombogle:
This fixes Writing System bug when exporting a session to FLEx

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/73)


hatton:
Tom can you tell us how all these .net changes were related to the bug? e.g. a good pr description might be

> Update newtonsoft-json to fix json which was causing a parsing error while trying to read FLEx ____

hatton:
Ah, the commit message has roughly that.

PR #77, opened 2018-10-26
Update package versions

tombogle:
Fixed mismatch in version between tests and production DLL that was causing the wrong version of Newtonsoft.json.dll to get installed.
Also upgraded (minor version) to newest DesktopAnalytics.


tombogle:
Oops. I just noticed that this also includes a previous commit to update a batch file (see PR #75). In a perfect world, that PR would be merged before this one, but whatever. The batch file is just for use by a dev on a local machine and is not involved in the build system at all.

PR #79, opened 2018-11-27
Repair things so unit tests pass after recent SLDR change

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/79)


JohnThomson:
The other approach finally worked, so this is not needed.

PR #81, opened 2018-11-29
Don't store roles as part of the file Participants list

JohnThomson:
In general, attempts to remove use of the Participants list except to maintain it for backwards compatibility. Instead, we derive the data from Contributions.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/81)

### DIFF
--- a/src/SayMore/Model/ArchivingHelper.cs
+++ b/src/SayMore/Model/ArchivingHelper.cs
@@ -17,6 +17,7 @@ using SIL.Archiving.Generic;
 using SIL.Archiving.IMDI;
 using SayMore.Properties;
 using SIL.Archiving.IMDI.Lists;
+using SIL.Windows.Forms.ClearShare;
 using SIL.WritingSystems;
 using Application = System.Windows.Forms.Application;
 
@@ -270,15 +271,13 @@ namespace SayMore.Model
 			// actors
 			var actors = new ArchivingActorCollection();
 			var persons = saymoreSession.GetAllPersonsInSession();
-			var allParticipantsWithRoles = saymoreSession.MetaDataFile.MetaDataFieldValues.Count > 0?
-				(from p in saymoreSession.MetaDataFile.GetStringValue(SessionFileType.kParticipantsFieldName, string.Empty).Split(FieldInstance.kDefaultMultiValueDelimiter)
-											where p.Trim() != string.Empty
-											select p.Trim()).ToList()
-				: new List<string>();
+			var contributions =
+				saymoreSession.MetaDataFile.GetValue(SessionFileType.kContributionsFieldName, null) as ContributionCollection;
+
 			foreach (var person in persons)
 			{
 
-				var actor = InitializeActor(model, person, sessionDateTime, GetRole(person.Id, actors, allParticipantsWithRoles));
+				var actor = InitializeActor(model, person, sessionDateTime, GetRole(person.Id, actors, contributions));
 
 				// do this to get the ISO3 codes for the languages because they are not in saymore
 				var language = GetOneLanguage(ForceIso639ThreeChar(person.MetaDataFile.GetStringValue("primaryLanguage", null)));
@@ -397,14 +396,9 @@ namespace SayMore.Model
 			return analysisLanguage;
 		}
 
-		private static string GetRole(string personId, ArchivingActorCollection actors, System.Collections.Generic.IEnumerable<string> allParticipantsWithRoles)
+		private static string GetRole(string personId, ArchivingActorCollection actors, ContributionCollection contributions)
 		{
-			return (from p in allParticipantsWithRoles
-				where p.StartsWith(personId + " (")
-				let openParen = p.IndexOf("(", StringComparison.InvariantCulture)
-				let closeParen = p.IndexOf(")", StringComparison.InvariantCulture)
-				select p.Substring(openParen + 1, closeParen - openParen - 1))
-				.FirstOrDefault(role => !PersonHasRole(personId, actors, role));
+			return contributions?.Where(c => c.ContributorName == personId).Select(c => c.Role).FirstOrDefault()?.Name;
 		}
 
 		private static bool PersonHasRole(string personId, ArchivingActorCollection actors, string role)
--- a/src/SayMore/Model/Files/FileType.cs
+++ b/src/SayMore/Model/Files/FileType.cs
@@ -190,7 +190,7 @@ namespace SayMore.Model.Files
 		/// ------------------------------------------------------------------------------------
 		public virtual bool GetShowInPresetOptions(string key)
 		{
-			return (!GetIsCustomFieldId(key) && !GetIsReadonly(key) && key != "contributions");
+			return (!GetIsCustomFieldId(key) && !GetIsReadonly(key) && key != SessionFileType.kContributionsFieldName);
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -355,6 +355,7 @@ namespace SayMore.Model.Files
 		public const string kGenreFieldName = "genre";
 		public const string kParticipantsFieldName = "participants";
 		public const string kTitleFieldName = "title";
+		public const string kContributionsFieldName = "contributions";
 
 		// additional fields
 		public const string kCountryFieldName = XmlFileSerializer.kAdditionalFieldIdPrefix + "Location_Country";
@@ -434,7 +435,7 @@ namespace SayMore.Model.Files
 					kGenreFieldName,
 					kParticipantsFieldName,
 					kTitleFieldName,
-					"contributions",
+					kContributionsFieldName,
 					kInteractivityFieldName,
 					kInvolvementFieldName,
 					kCountryFieldName,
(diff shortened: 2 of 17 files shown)


tombogle:
If this is a fix for a Jira issue, it would be good to include the issue # in the PR (and/or the commit message).

JohnThomson:
John has been using Trello to track SayMore issues. This relates to https://trello.com/c/snu4qukY/26-new-saymore-throws-away-participants-from-previous-versions. Not sure how to use that in a commit title.

hatton:
@tombogle does this PR have your approval?

src/SayMore/UI/ComponentEditors/BindingHelper.cs
```diff
@@ -187,11 +187,59 @@ private void BindControl(Control ctrl)
...
 		}
 
+		HashSet<Control> _validateOnHideControls = new HashSet<Control>();
+
+		private bool _runningValidations;
+		// If the control or any of its parents gets hidden, e.g., by switching to another tab,
+		// perform validation, even if the new tab does not assign focus to something, which
+		// would normally trigger it. This ensures some important side effects of validation
+		// don't get skipped (e.g., updating a Person's name in Contributors).
+		void ValidateOnHide(Control start)
```
tombogle: The only reference to this new method is inside itself. Doesn't seem like it will ever get called.

src/SayMore/UI/ComponentEditors/ContributorsEditor.cs
```diff
@@ -234,16 +234,17 @@ private void SaveContributors()
 		{
 			string failureMessage;
 			_file.SetValue("contributions", _model.Contributions, out failureMessage);
-			_file.SetValue(SessionFileType.kParticipantsFieldName, GetParticipants(), out failureMessage);
+			_file.SetValue(SessionFileType.kParticipantsFieldName, GetParticipants(false), out failureMessage);
```
tombogle: This line is essentially a subset of the above information, right? I assume it is being stored for backwards compatibility. If so, probably needs a comment to that effect.

src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs
```diff
@@ -561,25 +571,34 @@ private void SetAccessCodeListAndValue()
...
+			// the contributions field, because we don't want the value really stored in participants,
+			// a list of names we maintain for backwards compatibility, to include the roles.
+			// Note that this value should only be displayed in the _participants control,
+			// not stored in the <participants> field (as was done briefly, breaking backwards
+			// compatibility and causing problems for other programs that use the file).
+			// (If the user could edit the field, we'd have to watch out that the binding helper didn't
+			// copy the edited names-with-roles into the file. But attempts to edit this field
+			// trigger a switch to the Contributors tab.)
+			var contributions = _file.GetValue(SessionFileType.kContributionsFieldName, null) as ContributionCollection;
+			var participantsWithRoles = string.Join("; ",
```
tombogle: I'm getting an exception here:
Msg: Value cannot be null.
Parameter name: source
Class: System.ArgumentNullException
Source: System.Core
Assembly: System.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089
Stack:    at System.Linq.Enumerable.Select[TSource,TResult](IEnumerable`1 source, Func`2 selector)
   at SayMore.UI.ComponentEditors.SessionBasicEditor.GetParticipantsWithRolesFromContributions() in C:\Projects\saymoreJT\src\SayMore\UI\ComponentEditors\SessionBasicEditor.cs:line 597
   at SayMore.UI.ComponentEditors.SessionBasicEditor.HandleBinderTranslateBoundValueBeingRetrieved(Object sender, TranslateBoundValueBeingRetrievedArgs args) in C:\Projects\saymoreJT\src\SayMore\UI\ComponentEditors\SessionBasicEditor.cs:line 581
   at SayMore.UI.ComponentEditors.BindingHelper.UpdateControlValueFromField(Control ctrl) in C:\Projects\saymoreJT\src\SayMore\UI\ComponentEditors\BindingHelper.cs:line 292
Thread: UI
Thread UI culture: en-US
Exception: System.ArgumentNullException

tombogle: If contributions is null, I think you just want to return an empty string.

src/SayMore/Model/Files/FileType.cs
```diff
@@ -355,6 +355,7 @@ public class SessionFileType : FileType
 		public const string kGenreFieldName = "genre";
 		public const string kParticipantsFieldName = "participants";
 		public const string kTitleFieldName = "title";
+		public const string kContributionsFieldName = "contributions";
```
tombogle: Looks like there are some more places in the code where we should be using this constant.

src/SayMore/Model/ArchivingHelper.cs
```diff
@@ -397,14 +396,13 @@ private static string GetAnalysisLanguageIdentifier(Project saymoreProject)
...
-		private static string GetRole(string personId, ArchivingActorCollection actors, System.Collections.Generic.IEnumerable<string> allParticipantsWithRoles)
+		private static string GetRole(string personId, ArchivingActorCollection actors, ContributionCollection contributions)
 		{
-			return (from p in allParticipantsWithRoles
-				where p.StartsWith(personId + " (")
-				let openParen = p.IndexOf("(", StringComparison.InvariantCulture)
-				let closeParen = p.IndexOf(")", StringComparison.InvariantCulture)
-				select p.Substring(openParen + 1, closeParen - openParen - 1))
-				.FirstOrDefault(role => !PersonHasRole(personId, actors, role));
+			if (contributions == null)
```
tombogle: This could be made more compact (and IMO, more readable) as follows:
return contributions?.Where(c => c.ContributorName == personId).Select(c => c.Role).FirstOrDefault()?.Name;

src/SayMore/Model/Session.cs
```diff
@@ -219,12 +219,12 @@ public override IEnumerable<ComponentRole> GetCompletedStages()
 		/// ------------------------------------------------------------------------------------
 		private bool GetShouldReportHaveConsent()
 		{
-			var allParticipants = MetaDataFile.GetStringValue(SessionFileType.kParticipantsFieldName, string.Empty);
-			var personNames = FieldInstance.GetMultipleValuesFromText(allParticipants).ToArray();
-			bool allParticipantsHaveConsent = personNames.Length > 0;
+			var contributions = MetaDataFile.GetValue(SessionFileType.kContributionsFieldName, null) as ContributionCollection;
+			var personNames = contributions.Select(c => c.ContributorName).ToArray();
```
tombogle: Possible null reference exception here.

src/SayMore/UI/ComponentEditors/ContributorsEditor.cs
```diff
@@ -252,12 +253,16 @@ private void SaveContributors()
...
 		{
 			string participants = string.Empty;
+			var names = new HashSet<string>();
 			foreach (Contribution contributor in _model.Contributions)
 			{
-				participants += contributor.ContributorName + " (" + contributor.Role.Name + "); ";
+				if (!withRoles && names.Contains(contributor.ContributorName))
+					continue;
+				names.Add(contributor.ContributorName);
+				participants += contributor.ContributorName + (withRoles? " (" + contributor.Role.Name + "); " : ";");
```
tombogle: Probably doesn't really matter, but the older version of SayMore did have a space following the semi-colon to separate the participants in the list.

src/SayMoreTests/model/ProjectTests.cs
```diff
@@ -331,6 +331,52 @@ public void SetContributorsListToSession_ContributorsInMetaAndSession_AllContrib
...
+			var sessionXml = new XmlDocument();
+			using (var xmlReader = XmlReader.Create(Path.Combine(SessionScenarioPath, "OldSession.session")))
+			{
+				sessionXml.Load(xmlReader);
+			}
+			var participants = sessionXml.SelectSingleNode("//participants").InnerText.Split(';').Select(x => x.Trim());
+			Assert.AreEqual(2, participants.Count(), "Expecting two participants (without roles): Iskender Demirel, Lahdo Agirman");
+			Assert.That(participants, Has.Member("Iskender Demirel"));
+			Assert.That(participants, Has.Member("Lahdo Agirman"));
+			Assert.AreEqual(2, sessionXml.SelectNodes("//contributions/contributor").Count, "Expecting two contributors: Iskender Demirel, Lahdo Agirman");
```
tombogle: Should we also assert that their role is participant?

tombogle: Maybe not necessary since next test does this.

tombogle:
___
*[src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs, line 597 at r1](https://reviewable.io/reviews/sillsdev/saymore/81#-LSVhoO4RlS2WaqvQJEM-r1-597:-LTZHY5081aJFBLGUsqG:beeizjg) ([raw file](https://github.com/sillsdev/saymore/blob/f3842ff8bfac2876b93ebdf8aa887f319f039d27/src/SayMore/UI/ComponentEditors/SessionBasicEditor.cs#L597)):*
<details><summary><i>Previously, JohnThomson (John Thomson) wrote…</i></summary><blockquote>

Done. (Though I'm surprised...I thought our migration code was making sure contributions is never missing.)
</blockquote></details>

It's always possible it was some weird/bad data condition that would never happen to a normal user.

PR #82, opened 2018-12-27
Don't crash on, e.g., <Person/>

hatton:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/82)

### DIFF
--- a/src/SayMore/Model/Files/XmlFileSerializer.cs
+++ b/src/SayMore/Model/Files/XmlFileSerializer.cs
@@ -220,25 +220,29 @@ namespace SayMore.Model.Files
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


tombogle:
Phil and I have both looked at this and don't see a way to make it crash (or even hiccup) on data like <person/>. If there is a Jira issue for this, it should be referenced by this PR. If not, and if there is a real problem, an issue should be created with clear steps to reproduce the crash. Otherwise, this PR should be closed.

hatton:
> Phil and I have both looked at this and don't see a way to make it crash (or even hiccup) on data 

I think that's an absurd measure of whether to merge something from 2 years ago, when my team was knee deep in this code.

PR #84, opened 2019-01-11
NuGet cleanup and minor build improvements

tombogle:
Removed some unnecessary VS2013 stuff for NuGet.
Improved proj file to allow it to be used for local test builds.
Made NuGet.targets more like the versions used in other projects.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/84)


tombogle:
JohnT, This is low priority (nothing's actually broken), so if you want to leave this until Phil Hopper comes on board and have him review it, that's fine.

PR #86, opened 2019-02-27
Export Careful Speech

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/86)


src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+		{
+			_textBoxOutput.Font = Program.DialogFont;
+			_labelOutputFile.Font = Program.DialogFont;
+			_labelOutputFileValue.Font = FontHelper.MakeFont(Program.DialogFont, FontStyle.Bold);
+			_labelDownloadNeeded.Font = Program.DialogFont;
+			_labelStatus.Font = _labelOutputFileValue.Font;
+		}
+
+		/// ------------------------------------------------------------------------------------
+		public static string GetFactoryConvertToH263Mp4ConversionName()
```
tombogle: We don't want to duplicate these strings. The ones here are not used anyway, but by having two different instances of the same localizable string, you set up a situation where they could end up getting out of sync and L10nSharp will choose one arbitrarily.

src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+			{
+				StartPosition = FormStartPosition.CenterScreen;
+				Settings.Default.ExportMediaDlg = FormSettings.Create(this);
+			}
+
+			InitializeFonts();
+
+			if (_viewModel.MediaInfo == null)
+			{
+				_labelStatus.Text = LocalizationManager.GetString("DialogBoxes.ConvertMediaDlg.InvalidMediaFile",
```
tombogle: Any localizable strings in this file that are exact copies of the original ones in ConvertMediaDlg need to be turned into statics there and accessed wherever needed to avoid duplicates. Otherwise, if there is a reasonable possibility that text of the strings could diverge (either in English or in a localized version), they need to be given unique IDs in both places.

src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+using L10NSharp.UI;
+using SIL.Reporting;
+using SIL.Windows.Forms;
+using SIL.Windows.Forms.PortableSettingsProvider;
+using SayMore.Media.FFmpeg;
+using SayMore.Properties;
+
+namespace SayMore.UI
+{
+	public partial class ExportMediaDlg : Form
```
tombogle: If it's not really feasible to have a single dialog that can be parameterized to provide the two different desired behaviors, we at least need a comment near the top of both versions pointing out that they have a lot of similar code that generally should be kept in sync so that improvements in the one are also made in the other.

src/SayMore/UI/ConvertMediaDlgViewModel.cs
```diff
@@ -38,6 +38,7 @@ public class ConvertMediaDlgViewModel
 		public ConvertMediaUIState ConversionState { get; private set; }
 		public MediaFileInfo MediaInfo { get; private set; }
 		public string OutputFileCreated { get; private set; }
+		public readonly string OutputFile;
```
tombogle: It doesn't look like this needs to be public. If it's going to be a private field, it should be named _outputFile. If it is needed publicly, it should be made a property (like the ones above it).

src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+			base.OnLoad(e);
+			UpdateDisplay();
+		}
+
+		/// ------------------------------------------------------------------------------------
+		protected override void OnClosing(CancelEventArgs e)
+		{
+			Program.ResumeBackgroundProcesses(true);
+
+			if ((_viewModel.ConversionState & ConvertMediaUIState.FinishedConverting) > 0)
```
tombogle: See above comments about DialogResult.

src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+using SayMore.Properties;
+
+namespace SayMore.UI
+{
+	public partial class ExportMediaDlg : Form
+	{
+		private readonly ConvertMediaDlgViewModel _viewModel;
+		private string _conversionInProgressStatusFormat;
+
+		public static DialogResult Show(string inputFile, string initialConversionName, string outputFile)
```
tombogle: The DialogResult here is not used.

src/SayMore/UI/ExportMediaDlg.cs
```diff
@@ -0,0 +1,260 @@
...
+			var viewModel = new ConvertMediaDlgViewModel(inputFile, initialConversionName, outputFile);
+			using (var dlg = new ExportMediaDlg(viewModel))
+				return dlg.ShowDialog();
+		}
+
+		/// ------------------------------------------------------------------------------------
+		public ExportMediaDlg()
+		{
+			InitializeComponent();
+			DialogResult = DialogResult.Cancel;
```
tombogle: DialogResult not really needed.

src/SayMore/Transcription/Model/Exporters/AnnotationAudioExporter.cs
```diff
@@ -45,57 +52,76 @@ public static void Export(string waveFileDirectoryName, string filenameFilter, s
...
 					}
 				}
 			}
-
-			return tempFileName;
+			catch (AnnotationAudioExporterException)
+			{
+				throw;
+			}
+			catch (FormatException)
```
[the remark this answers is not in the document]
phillip-hopper: I looked it up, and I think I'm doing it correctly. https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2200-rethrow-to-preserve-stack-details?view=vs-2017

src/SayMore/UI/ConvertMediaDlgViewModel.cs
phillip-hopper: Done.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Combined into one.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Combined into one.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Combined into one.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Combined into one.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Changed to be void.

src/SayMore/UI/ExportMediaDlg.cs
phillip-hopper: Combined into one.

src/SayMore/Transcription/Model/Exporters/AnnotationAudioExporter.cs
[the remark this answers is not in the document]
tombogle: In the case of an AnnotationAudioExporterException, you do correctly rethrow it. I was referring to the following block, where you catch a FormatException and then create a new FormatException and throw that one, which loses the call stack of the original one.

src/SayMore/UI/ConvertMediaDlgViewModel.cs
```diff
@@ -131,7 +141,13 @@ public string GetNewOutputFileName(bool returnFileNameOnly)
 			}
 
 			if (outputFile == null)
-				outputFile = GetNewOutputFileName(false);
+			{
+				if (!string.IsNullOrEmpty(_outputFile))
```
tombogle: This check here seems to be redundant, since GetNewOutputFilename already does this.

PR #90, opened 2019-03-27
Remove FFmpeg download dialog

phillip-hopper:
Remove the dialog and related code for downloading FFmpeg. Include FFmpeg in the installer now.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/90)


src/SayMore/UI/ConvertMediaDlg.cs
```diff
@@ -157,7 +148,6 @@ private void SetupForAutoRun()
 		/// ------------------------------------------------------------------------------------
 		private void HandleStringsLocalized()
 		{
-			_labelDownloadNeeded.Tag = _labelDownloadNeeded.Text;
 			UpdateDisplay();
```
tombogle: This method is probably no longer needed. We'd only need to call UpdateDisplay when the localized strings change if we're doing something related to the strings (e.g., setting a variable or property to the new value).

src/SayMore/UI/ConvertMediaDlgViewModel.cs
```diff
@@ -88,13 +89,6 @@ public string ConversionOutput
 			get { return _conversionOutput.ToString(); }
 		}
 
-		/// ------------------------------------------------------------------------------------
-		public void SetConversionStateBasedOnPresenceOfFfmpegForSayMore()
-		{
-			ConversionState = (FFmpegDownloadHelper.DoesFFmpegForSayMoreExist ?
```
tombogle: ConvertMediaUIState.FFmpegDownloadNeeded (and the code that uses it) can go away completely.

src/SayMore/UI/ConvertMediaDlg.cs
phillip-hopper: Done.

src/SayMore/UI/ConvertMediaDlgViewModel.cs
phillip-hopper: Done.

phillip-hopper:
I removed the `license.txt` file from the `FFmpeg` directory. That file appears to be outdated (2007) compared to the website.

PR #93, opened 2019-05-08
Fix crashes when the last session or person are deleted

phillip-hopper:
Was getting Directory not found. After that was fixed, started getting object not set.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/93)

### DIFF
--- a/src/SayMore/Model/Person.cs
+++ b/src/SayMore/Model/Person.cs
@@ -24,6 +24,7 @@ namespace SayMore.Model
 		{
 			get
 			{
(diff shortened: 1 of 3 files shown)


src/SayMore/Model/Person.cs
```diff
@@ -24,6 +24,9 @@ public override string UiId
 		{
 			get
 			{
+				if (MetaDataFile == null)
```
tombogle: Might be good to add a comment explaining why this might be null.

PR #100, opened 2019-07-17
Detect, stop and restart file sync clients

phillip-hopper:
Includes Dropbox, Google Drive and Microsoft OneDrive

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/100)

### DIFF
--- a/src/SayMore/Utilities/FileSyncHelper.cs
+++ b/src/SayMore/Utilities/FileSyncHelper.cs
@@ -30,41 +30,43 @@ namespace SayMore.Utilities
 		/// Checks if the file is being synced, and prompts user to disable syncing if it is.
 		/// </summary>
 		/// <param name="filePath"></param>
-		public static void PromptToStopSync(string filePath)
+		/// <returns>If a client is stopped, the client is returned, otherwise SyncClient.None</returns>
+		public static SyncClient PromptToStopSync(string filePath)
 		{
 			var client = IsSynched(filePath);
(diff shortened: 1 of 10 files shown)


src/SayMore/Utilities/FileSyncHelper.cs
```diff
@@ -0,0 +1,322 @@
...
+					{
+						KillProcess(kGoogleDriveProcess);
+						_stoppedGoogleDrive = true;
+					}
+					break;
+
+				case SyncClient.OneDrive:
+					if (!_stoppedOneDrive)
+					{
+						// OneDrive has an off switch
```
tombogle: This code block is identical to the code for turning it back on, except for the command line parameter. Make a method.

src/SayMore/Utilities/FileSyncHelper.cs
```diff
@@ -0,0 +1,322 @@
...
+		}
+
+		/// <summary>
+		/// First checks if the file is in a potentially synced directory,
+		/// then if it is, checks if the sync program is active.
+		/// </summary>
+		/// <param name="pathParts"></param>
+		/// <param name="client"></param>
+		/// <returns></returns>
+		private static bool IsThisClient(string[] pathParts, SyncClient client)
```
tombogle: Might want to come up with a name that more accurately suggests that its job is also to see if the client is active.

PR #102, opened 2019-07-17
SP-2118: FieldWorks installs an incompatible version

phillip-hopper:
The version of Microsoft.Extensions.DependencyModel.dll shipping with Fieldworks is not compatible with SayMore

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/102)


tombogle:
Eberhard looked at this and concluded:
It's strange what's going on: icu.net 2.5 depends on Microsoft.Extensions.DependencyModel (which saymore includes as nuget package). So far the installer didn't install that. Furthermore, the released version 3.1.5 as well as the version from SP-2118 (3.2.2248) came with an older icu.net version (2.1) which didn't have this dependency. I'm not sure why after the installation of FW saymore suddenly uses icu.net 2.5. But in any case, the PR looks fine: icu.net 2.5 has that dependency, so saymore should install the two assemblies.

tombogle:
FWIW, HearThis (which still needs to be tested in this scenario) does get ICU via NuGet, rather than from TeamCity.

PR #104, opened 2019-08-14
SP-2112: Pressing Enter key in should tab to next control

phillip-hopper:
Pressing the Enter key in single line control should tab to next control rather than playing an "error sound."

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/104)

### DIFF
--- a/src/SayMore/UI/ProjectWindow/ProjectWindow.cs
+++ b/src/SayMore/UI/ProjectWindow/ProjectWindow.cs
@@ -45,6 +45,8 @@ namespace SayMore.UI.ProjectWindow
 		private MPlayerDebuggingOutputWindow _outputDebuggingWindow;
 		private string _titleFmt;
 
+		private static string[] _keyDownTypes = {"TextBox", "ComboBox", "DatePicker"};
+
 		public bool UserWantsToOpenADifferentProject { get; set; }
 
 		/// ------------------------------------------------------------------------------------
@@ -61,9 +63,51 @@ namespace SayMore.UI.ProjectWindow
 			ExceptionHandler.AddDelegate(AudioUtils.HandleGlobalNAudioException);
 
 			InitializeComponent();
+			// ReSharper disable once VirtualMemberCallInConstructor
 			_titleFmt = Text;
 			_menuShowMPlayerDebugWindow.Tag = _menuProject.DropDownItems.IndexOf(_menuShowMPlayerDebugWindow);
 			_menuProject.DropDownItems.Remove(_menuShowMPlayerDebugWindow);
+
+			KeyPreview = true;
+			KeyDown += Project_KeyDown;
+		}
+
+		private static Control FindFocusedControl(Control control)
+		{
+			var container = control as IContainerControl;
+			while (container != null)
+			{
+				control = container.ActiveControl;
+				container = control as IContainerControl;
+			}
+			return control;
+		}
+
+		private void Project_KeyDown(object sender, KeyEventArgs e)
+		{
+			var control = FindFocusedControl((Control)sender);
+
+			if (!_keyDownTypes.Contains(control.GetType().Name)) return;
+			if (e.KeyCode != Keys.Return) return;
+
+			// skip multiline text boxes
+			if (control is TextBox txt)
+			{
+				if (txt.Multiline) return;
+			}
+			
+			// prevent the ding
+			e.SuppressKeyPress = true;
+
+			// go to the next control in the tab order
+			var next = GetNextControl(control, true);
+			while (next != null)
+			{
+				if (_keyDownTypes.Contains(next.GetType().Name)) break;
+				next = GetNextControl(next, true);
+			}
+
+			next?.Focus();
 		}
 
 		/// ------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/SayMore/UI/ProjectWindow/ProjectWindow.cs
```diff
@@ -61,9 +63,51 @@ private ProjectWindow()
 			ExceptionHandler.AddDelegate(AudioUtils.HandleGlobalNAudioException);
 
 			InitializeComponent();
+			// ReSharper disable once VirtualMemberCallInConstructor
 			_titleFmt = Text;
 			_menuShowMPlayerDebugWindow.Tag = _menuProject.DropDownItems.IndexOf(_menuShowMPlayerDebugWindow);
 			_menuProject.DropDownItems.Remove(_menuShowMPlayerDebugWindow);
+
+			KeyPreview = true;
+			KeyDown += Project_KeyDown;
```
tombogle: I think you can just override OnKeyDown, rather than hooking up a handler. Better yet, I think you could leave KeyPreview false and just override ProcessDialogKey or ProcessCmdKey. (According to [this](https://stackoverflow.com/questions/2386695/disadvantage-of-setting-form-keypreview-true/2386844#2386844), KeyPreview is a bit of an anachronism held over from the days of VB6.)

src/SayMore/UI/ProjectWindow/ProjectWindow.cs
```diff
@@ -61,9 +63,51 @@ private ProjectWindow()
...
+			}
+			
+			// prevent the ding
+			e.SuppressKeyPress = true;
+
+			// go to the next control in the tab order
+			var next = GetNextControl(control, true);
+			while (next != null)
+			{
+				if (_keyDownTypes.Contains(next.GetType().Name)) break;
```
tombogle: I think what you want is to use Control.SelectNextControl.
It's not clear to me why these three types of controls are the only legitimate landing sites. For example, if there were a datagridview control next in line, wouldn't it be legit to land there? If this is just to prevent landing inside a container, maybe the logic should be checking to see if the control is an IContainerControl. (Perhaps a comment is needed to say what kinds of controls you're trying to avoid and/or why. Or maybe this needs to be a separate list.)

src/SayMore/UI/ProjectWindow/ProjectWindow.cs
```diff
@@ -45,6 +45,8 @@ public partial class ProjectWindow : Form
 		private MPlayerDebuggingOutputWindow _outputDebuggingWindow;
 		private string _titleFmt;
 
+		private static string[] _keyDownTypes = {"TextBox", "ComboBox", "DatePicker"};
```
tombogle: the name _keyDownTypes isn't very clear.  Perhaps you can come up with a name that better conveys the intended meaning. A better approach might be to add a small static class with an extension method:
`static bool ShouldTreatEnterAsTab(this Control ctrl) // alternative name: ShouldEnterAdvanceToNextControl`
Your implementation could then just ask is ctrl TextBox, ComboBox, etc.. This would be more efficient than doing a string comparison on type name and would nicely extend to derived types. The implementation could also nicely deal with the special case of a Multiline text box (without having to check its type twice).

src/SayMore/UI/ProjectWindow/ProjectWindow.cs
```diff
@@ -61,9 +63,51 @@ private ProjectWindow()
...
+			if (e.KeyCode != Keys.Return) return;
+
+			// skip multiline text boxes
+			if (control is TextBox txt)
+			{
+				if (txt.Multiline) return;
+			}
+			
+			// prevent the ding
+			e.SuppressKeyPress = true;
```
tombogle: If you use ProcessDialogKey or ProcessCmdKey, you can just return `true`, and this won't be needed. (Even doing it this way, I would think you could set e.Handled = true, but maybe that doesn't work, since this is just handling the key down event.)

tombogle:
___
*[src/SayMore/UI/ProjectWindow/ProjectWindow.cs, line 106 at r1](https://reviewable.io/reviews/sillsdev/saymore/104#-LmFAF0JBwdi__9Yw-n6-r1-106:-LpoZiKT-gD34XFJNba8:b-c38gs6) ([raw file](https://github.com/sillsdev/saymore/blob/04209712453992c539b31730ce90231c3236c8dc/src/SayMore/UI/ProjectWindow/ProjectWindow.cs#L106)):*

Also don't want to land on buttons or links.
</blockquote></details>

Please make that explicit in a comment. Also, consider whether "wrapping" is desirable. SelectNextControl allows for wrapping. Unfortunately, GetNextControl doesn't.

tombogle:
___
*[src/SayMore/Utilities/ExtensionMethods.cs, line 52 at r3](https://reviewable.io/reviews/sillsdev/saymore/104#-Lpoa2xeAN56AA8vQZyx:-Lpoa2xeAN56AA8vQZyy:bitap21) ([raw file](https://github.com/sillsdev/saymore/blob/3dd9ae4c3ec32959f05af3e2c25fc6c5868e35b3/src/SayMore/Utilities/ExtensionMethods.cs#L52)):*
> ```Smalltalk
> 		/// <returns></returns>
> 		public static bool ShouldTabToMe(this Control ctrl)
> 		{
> ```

Not a huge, deal but this would be more compact:
 			return ctrl is ComboBox ||
				ctrl is DatePicker ||
				ctrl is TextBox ||
				ctrl is DataGridView;

PR #105, opened 2019-08-28
SP-2114: Update data on the Project Progress tab

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/105)

### DIFF
--- a/src/SayMore/Model/Files/DataGathering/AudioVideoDataGatherer.cs
+++ b/src/SayMore/Model/Files/DataGathering/AudioVideoDataGatherer.cs
@@ -1,8 +1,6 @@
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using SayMore.Media;
-using SayMore.Properties;
 
 namespace SayMore.Model.Files.DataGathering
 {
@@ -21,26 +19,6 @@ namespace SayMore.Model.Files.DataGathering
 			base(rootDirectoryPath, allFileTypes.Where(t => t.IsAudioOrVideo),
 				MediaFileInfo.GetInfo)
 		{
-		}
-
-		/// ------------------------------------------------------------------------------------
-		protected override bool GetDoIncludeFile(string path)
-		{
-			// Don't collect info. on oral annotation segment files.
-			var isOralAnnotationSegmentFile =
-				Path.GetDirectoryName(path.ToLower()).EndsWith(Settings.Default.OralAnnotationsFolderSuffix.ToLower());
-
-			return (!isOralAnnotationSegmentFile && base.GetDoIncludeFile(path));
-		}
-
-		/// ------------------------------------------------------------------------------------
-		public void ProcessThisFile(string fileName)
-		{
-			if (!GetDoIncludeFile(fileName))
-				return;
-			SuspendProcessing();
-			CollectDataForFile(fileName);
-			ResumeProcessing(true);
 		}
 	}
 }
--- a/src/SayMore/UI/Overview/Statistics/StatisticsViewModel.cs
+++ b/src/SayMore/UI/Overview/Statistics/StatisticsViewModel.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Collections.Generic;
 using L10NSharp;
+using SayMore.Media;
 using SayMore.Model;
 using SayMore.Model.Files;
 using SayMore.Model.Files.DataGathering;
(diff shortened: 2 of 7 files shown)


src/SayMore/Model/Files/DataGathering/AudioVideoDataGatherer.cs
```diff
@@ -22,25 +20,5 @@ public class AudioVideoDataGatherer :
...
-		{
-			// Don't collect info. on oral annotation segment files.
-			var isOralAnnotationSegmentFile =
-				Path.GetDirectoryName(path.ToLower()).EndsWith(Settings.Default.OralAnnotationsFolderSuffix.ToLower());
-
-			return (!isOralAnnotationSegmentFile && base.GetDoIncludeFile(path));
-		}
-
-		/// ------------------------------------------------------------------------------------
-		public void ProcessThisFile(string fileName)
```
tombogle: As near as I can tell, this was not even used before.

src/SayMore/UI/Overview/Statistics/StatisticsViewModel.cs
```diff
@@ -101,20 +102,19 @@ public IEnumerable<ComponentRoleStatistics> GetComponentRoleStatisticsPairs()
...
-			long bytes = 0;
-			foreach (var info in _backgroundStatisticsGather.GetAllFileData())
-			{
-				if (role.IsMatch(info.MediaFilePath))
-					bytes += info.LengthInBytes;
-			}
-
-			return bytes;
+			// do not include the generated oral annotation file in the calculations
+			return _backgroundStatisticsGather.GetAllFileData()
```
tombogle: Seems that this method and GetRecordingDurations need identical logic to get the file data over which they are to work. Therefore, this should be pulled out into a private helper method they can share, rather than duplicating the logic. Also, for improved efficiency, I think the two Where clauses should be joined into one.

src/SayMoreTests/ApplicationContainerTests.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace SayMoreTests
+{
+	[TestFixture]
+	public class ApplicationContainerTests
+	{
+		[Test]
+		public void ComponentRoles_ContainsOralTranslationRole_HasCorrectRenamingSuffix()
+		{
+			// get list of all available roles
+			var roles = ApplicationContainer.ComponentRoles.ToList();
```
tombogle: No need to get these into a list. Also, since we only expect one match, instead of FirstOrDefault, call Single(r => r.Id == ComponentRole.kOralTranslationComponentRoleId) directly on ApplicationContainer.ComponentRoles.

src/SayMoreTests/ApplicationContainerTests.cs
```diff
@@ -0,0 +1,27 @@
...
+using SayMore.Model.Files;
+using SayMore.Properties;
+
+namespace SayMoreTests
+{
+	[TestFixture]
+	public class ApplicationContainerTests
+	{
+		[Test]
+		public void ComponentRoles_ContainsOralTranslationRole_HasCorrectRenamingSuffix()
```
tombogle: Normal three-part naming convention is <method or property under test>_<condition>_<expected result>

src/SayMoreTests/ApplicationContainerTests.cs
```diff
@@ -0,0 +1,27 @@
...
+using SayMore;
+using SayMore.Model.Files;
+using SayMore.Properties;
+
+namespace SayMoreTests
+{
+	[TestFixture]
+	public class ApplicationContainerTests
+	{
+		[Test]
```
tombogle: Could probably use TestCase and go ahead and test all relevant ComponentRoles that follow this pattern.

src/SayMore/UI/Overview/Statistics/StatisticsViewModel.cs
phillip-hopper: Done.

src/SayMoreTests/ApplicationContainerTests.cs
phillip-hopper: Done.

src/SayMoreTests/ApplicationContainerTests.cs
phillip-hopper: Done

tombogle:
___
*[src/SayMoreTests/ApplicationContainerTests.cs, line 12 at r1](https://reviewable.io/reviews/sillsdev/saymore/105#-LnNoXzyoyCAGWjkkT5a-r1-12:-LqGyQb92WqFyzeh7qmP:b-lj8nq8) ([raw file](https://github.com/sillsdev/saymore/blob/04e572c475a86a424932c7c448c1b3f9936d2645/src/SayMoreTests/ApplicationContainerTests.cs#L12)):*

Done
</blockquote></details>

Since the ExpectedResult is always true, you could eliminate the third parameter, make the method void, and just Assert.IsTrue instead of returning the result of the StartsWith. It probably won't make much of a difference in the way the results of the test are presented, but it seems a little more straightforward to me.

tombogle:
___
*[src/SayMoreTests/ApplicationContainerTests.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/saymore/105#-LqGzHMb0XpmRsnETpnL:-LqGzHMc5vWqv3-vQQ2L:b-rqdygd) ([raw file](https://github.com/sillsdev/saymore/blob/c581ed761657b6a7c3bfb9ddcf1353d7c7257b7e/src/SayMoreTests/ApplicationContainerTests.cs#L17)):*
> ```Smalltalk
> 			// get the oral translation role
> 			var role = ApplicationContainer.ComponentRoles.Single(r => r.Id == roleId);
> 			Assert.NotNull(role);
> ```

Asserting NotNull is unnecessary here. Single will throw an exception if one is not found.

tombogle:
___
*[src/SayMore/Model/Files/DataGathering/AudioVideoDataGatherer.cs, line 25 at r2](https://reviewable.io/reviews/sillsdev/saymore/105#-LqH0FJzCj0woFXs1SZg:-LqH0FJzCj0woFXs1SZh:b9ehan4) ([raw file](https://github.com/sillsdev/saymore/blob/3911bc5ac939ca4b459cad5375147db48bc64315/src/SayMore/Model/Files/DataGathering/AudioVideoDataGatherer.cs#L25)):*
> 		protected override bool GetDoIncludeFile(string path)
> 		public void ProcessThisFile(string fileName)

Were these two methods just not being used at all?

PR #106, opened 2019-10-02
SP-2135: Do not throw fatal exception if slider value is out of range

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/106)

### DIFF
--- a/src/SayMore/UI/LowLevelControls/Slider.cs
+++ b/src/SayMore/UI/LowLevelControls/Slider.cs
@@ -2,7 +2,10 @@ using System;
 using System.ComponentModel;
 using System.Drawing;
 using System.Drawing.Drawing2D;
+using System.Globalization;
 using System.Windows.Forms;
+using L10NSharp;
+using SIL.Reporting;
 
 namespace SayMore.UI.LowLevelControls
 {
@@ -170,8 +173,19 @@ namespace SayMore.UI.LowLevelControls
 		{
 			value = (float)Math.Round(value, 4);
 
-			if (value < 0 || value > Maximum)
-				throw new ArgumentOutOfRangeException("value");
+			if (value < 0)
+			{
+				var msg = LocalizationManager.GetString("CommonToMultipleViews.Slider.ValueIsNegative", "Attempted to set slider to a negative value: {0}.");
+				ErrorReport.NotifyUserOfProblem(msg, value.ToString(CultureInfo.InvariantCulture));
+				return false;
+			}
+
+			if (value > Maximum)
+			{
+				var msg = LocalizationManager.GetString("CommonToMultipleViews.Slider.ValueGreaterThanMaximum", "Attempted to set slider to a value ({0}) which is greater than the maximum ({1}).");
+				ErrorReport.NotifyUserOfProblem(msg, value.ToString(CultureInfo.InvariantCulture), Maximum.ToString(CultureInfo.InvariantCulture));
+				return false;
+			}
 
 			if (value == _value && IsHandleCreated)
 				return false;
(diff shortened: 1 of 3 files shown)


src/SayMore/UI/LowLevelControls/Slider.cs
```diff
@@ -170,8 +173,19 @@ public bool SetValueWithoutEvent(float value)
 		{
 			value = (float)Math.Round(value, 4);
 
-			if (value < 0 || value > Maximum)
-				throw new ArgumentOutOfRangeException("value");
+			if (value < 0)
+			{
+				var msg = LocalizationManager.GetString("CommonToMultipleViews.Slider.ValueIsNegative", "Attempted to set slider to a negative value: {0}.");
```
tombogle: Since the user won't know what a "slider" is, could we possibly figure out a way to provide a more user-friendly name? One possibility might be to use "the {0} slider", where param 0 is **nameof(this)**, but depending on how the sliders are named, that might not be any better. If we can come up with a workable solution, do the same for the other message below.

phillip-hopper: Would it be better to just log the warning and not bother the user?

tombogle:
___
*[src/SayMore/UI/LowLevelControls/Slider.cs, line 178 at r1](https://reviewable.io/reviews/sillsdev/saymore/106#-LqCANHJkmU_j23e-P3d-r1-178:-M3I2myg1zMan5Dttn_E:bmzfcqt) ([raw file](https://github.com/sillsdev/saymore/blob/cf509d2d70ef20153f1d39f0628e8d99c10d34d3/src/SayMore/UI/LowLevelControls/Slider.cs#L178)):*

Would it be better to just log the warning and not bother the user?
</blockquote></details>

Probably. I can't remember when/why/how this was occurring. But if you think that's the right thing to do, go ahead and add the Logger statements.

src/SayMore/UI/LowLevelControls/Slider.cs
```diff
@@ -170,8 +173,11 @@ public bool SetValueWithoutEvent(float value)
 		{
 			value = (float)Math.Round(value, 4);
 
-			if (value < 0 || value > Maximum)
-				throw new ArgumentOutOfRangeException("value");
+			if (value < 0)
+				value = 0;
```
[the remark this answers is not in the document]
hatton: Tom I think you might like to have Sentry wired up to SayMore. LSDEV is running our own instance now. I've found with lameta that it's nice for *me* to get notice of errors that realistically, are meaningless to users and which they wouldn't tell me about.

I'll make an organization in case you want to go this direction.

See https://sentry.keyman.com/saymore/c/getting-started/csharp/

tombogle:
___
*[src/SayMore/UI/LowLevelControls/Slider.cs, line 177 at r2](https://reviewable.io/reviews/sillsdev/saymore/106#-LqCANHJkmU_j23e-P3d-r2-177:-M5cu1H21IzS_smTVmC4:b-nr6pqu) ([raw file](https://github.com/sillsdev/saymore/blob/104deb387fc7c2930143df3f414f816352bf9f08/src/SayMore/UI/LowLevelControls/Slider.cs#L177)):*

Tom I think you might like to have Sentry wired up to SayMore. LSDEV is running our own instance now. I've found with lameta that it's nice for *me* to get notice of errors that realistically, are meaningless to users and which they wouldn't tell me about.

I'll make an organization in case you want to go this direction.

See https://sentry.keyman.com/saymore/c/getting-started/csharp/
</blockquote></details>

Phil, why don't you look into this? I think this is probably the way to go.

tombogle:
After discussing this with Bryan, I don't think we should hold up the release for this. Let's merge as is, and then do a follow on task that can be released later to hook up Sentry so we can try to find out what's happening if this error ever occurs again.

PR #110, opened 2020-02-26
SP-1799: Object reference not set to an instance of an object

phillip-hopper:
Not able to reproduce, so I added a check for null and write to log if the value is null.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/110)

### DIFF
--- a/src/SayMore/MediaUtils/Audio/WavePainting/WavePainterBasic.cs
+++ b/src/SayMore/MediaUtils/Audio/WavePainting/WavePainterBasic.cs
@@ -494,6 +494,12 @@ namespace SayMore.Media.Audio
 				{
 					var sampleAmplitudes = _samplesToDraw[sampleToDraw, channel];
 
+					if (sampleAmplitudes == null)
+					{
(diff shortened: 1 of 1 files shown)


src/SayMore/MediaUtils/Audio/WavePainting/WavePainterBasic.cs
```diff
@@ -494,6 +494,12 @@ protected virtual void DrawWave(PaintEventArgs e, Rectangle rc)
 				{
 					var sampleAmplitudes = _samplesToDraw[sampleToDraw, channel];
 
+					if (sampleAmplitudes == null)
+					{
+						Logger.WriteEvent("'sampleAmplitudes' was null; 'sampleToDraw' or 'channel' possibly off by one. (See SP-1799)");
```
tombogle: Let's go ahead and include the values of the two indexes and the dimensions of the array also so we can see whether we're at a boundary case.

phillip-hopper: Done.

PR #112, opened 2020-03-25
SP-2159: Use NuGet for all dependencies

phillip-hopper:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/112)


tombogle:
___
*[src/SayMoreTests/packages.config, line 26 at r1](https://reviewable.io/reviews/sillsdev/saymore/112#-M3I8rKN1lY7mRcRTspe:-M3I8rKN1lY7mRcRTspf:b-vptjm8) ([raw file](https://github.com/sillsdev/saymore/blob/a7e483210c8d00b10acc7cc2c216f09d52fd58be/src/SayMoreTests/packages.config#L26)):*
> ```XML
>   <package id="NUnit.ConsoleRunner" version="3.11.1" targetFramework="net461" />
>   <package id="NUnit3TestAdapter" version="3.16.1" targetFramework="net461" developmentDependency="true" />
>   <package id="PangoSharp-signed" version="3.22.24.37" targetFramework="net461" />
> ```

I think that a bunch of these files are not actually required in the Windows Installer. For example, both Glyssen and HearThis use packages that depend on PangoSharp, but neither of their installers include it. I think what I did was tried the instller as it was and then added only the handful of files that actually resulted in load errors.

tombogle:
___
*[src/SayMore/packages.config, line 25 at r1](https://reviewable.io/reviews/sillsdev/saymore/112#-M3I9puF87P47dIpx6Gk:-M3I9puF87P47dIpx6Gl:bynrcdk) ([raw file](https://github.com/sillsdev/saymore/blob/a7e483210c8d00b10acc7cc2c216f09d52fd58be/src/SayMore/packages.config#L25)):*
> ```XML
>   <package id="Microsoft.Extensions.DependencyModel" version="3.1.2" targetFramework="net461" />
>   <package id="Mono.Posix" version="5.4.0.201" targetFramework="net461" />
>   <package id="NAudio" version="1.10.0" targetFramework="net461" />
> ```

SIL.Media references the "official" NAudio package, but SayMore needs our version that supports large WAV files. I think we're going to have to finish creating the nuget package for our version of NAudio and then change SIL.Media to reference it instead. (Or it might be possible to have SILMedia continue to depend on the official one and have SayMore install both (with different names. But I'm not sure there is any compelling reason to do that. Might need some discussion in the larger group.)

tombogle:
___
*[src/SayMore/SayMore.csproj, line 40 at r1](https://reviewable.io/reviews/sillsdev/saymore/112#-M3IAycJ-6XFr016nsJ1:-M3IAycJ-6XFr016nsJ2:b-9exttz) ([raw file](https://github.com/sillsdev/saymore/blob/a7e483210c8d00b10acc7cc2c216f09d52fd58be/src/SayMore/SayMore.csproj#L40)):*
> ```XML
>     </TargetFrameworkProfile>
>     <NuGetPackageImportStamp>
>     </NuGetPackageImportStamp>
> ```

If I understand it correctly, these lines are cruft and can be removed.

tombogle:
___
*[src/Installer/Installer.wxs, line 234 at r1](https://reviewable.io/reviews/sillsdev/saymore/112#-M3IBfoY5YOn5bAqgJvE:-M3IBfoY5YOn5bAqgJvF:b-vptjm8) ([raw file](https://github.com/sillsdev/saymore/blob/a7e483210c8d00b10acc7cc2c216f09d52fd58be/src/Installer/Installer.wxs#L234)):*
> ```XML
> 				<File Id="Newtonsoft.Json.dll" Name="Newtonsoft.Json.dll" KeyPath="yes" Source="..\..\output\release\Newtonsoft.Json.dll" />
> 			</Component>
> 			<Component Id="PangoSharp.dll" Guid="{3EA2217F-6FD9-487E-BAFA-A579955CD67C}">
> ```

I think that a bunch of these files are not actually required in the Windows Installer. For example, both Glyssen and HearThis use packages that depend on PangoSharp, but neither of their installers include it. I think what I did was tried the instller as it was and then added only the handful of files that actually resulted in load errors.

tombogle:
___
*[src/SayMoreTests/packages.config, line 20 at r1](https://reviewable.io/reviews/sillsdev/saymore/112#-M3IGKOo7MgQqoE-ef2t:-M3IGKOp0PeOmST3309k:bqc9jxg) ([raw file](https://github.com/sillsdev/saymore/blob/a7e483210c8d00b10acc7cc2c216f09d52fd58be/src/SayMoreTests/packages.config#L20)):*
> ```XML
>   <package id="Microsoft.Extensions.DependencyModel" version="3.1.2" targetFramework="net461" />
>   <package id="Mono.Posix" version="5.4.0.201" targetFramework="net461" />
>   <package id="NAudio" version="1.10.0" targetFramework="net461" />
> ```

See other comment regarding NAudio conflict.

tombogle:
Since it looks like the officiial release of NAudio now correctly handles large WAV files, we can just reference that package and avoid this conflict. If possible, it would be good to try to confirm that the deadlocking issue is also fixed. (I looked at the code in the official repo, and it had changed considerably. I wasn't convinced that the changes would avoid a deadlock, but it's at least possible.) Once we're ready to move forward with the official one, we can pause and hide the TC builds and maybe note that our fork of the repo is now inactive.

PR #114, opened 2020-08-09
SP-1000: Reduce minimum segment length from 0.5 seconds to 0.46

phillip-hopper:
Tester requested the minimum be reduced more. 0.46 was the smallest I could get and still have the UI display correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/114)

### DIFF
--- a/src/SayMore/Transcription/UI/SegmentingAndRecording/OralAnnotationRecorderDlgBase.cs
+++ b/src/SayMore/Transcription/UI/SegmentingAndRecording/OralAnnotationRecorderDlgBase.cs
@@ -8,7 +8,6 @@ using DesktopAnalytics;
 using L10NSharp;
 using L10NSharp.TMXUtils;
 using L10NSharp.UI;
-using SIL.Media.Naudio;
 using SIL.Media.Naudio.UI;
 using SIL.Reporting;
 using SIL.Windows.Forms;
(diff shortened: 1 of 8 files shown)


src/SayMore/Transcription/UI/SegmentingAndRecording/OralAnnotationRecorderDlgBase.cs
```diff
@@ -1869,8 +1878,12 @@ private Rectangle PlayAnnotationButtonRectangle
 
 				// SP-1000: Reduce minimum segment size
 				// Reducing the minimum segment size below 850 ms requires repositioning the Play Annotation button
+				if (rc.Width < 45)
+					return new Rectangle(rc.Left + 3, rc.Bottom - 40 - _playButtonSize.Height,
```
tombogle, line 1882: I believe the magic numbers "45", "80", and "40" based on the size of the two buttons? If so, could we use those sizes  instead? Also, to ensure the two buttons stay aligned (i.e, that consistent margins are used), I think it would probably be good to create a method that returns a rectangle, given a bounding rectangle (rc), an offset from bottom, and a button size. It could even include the logic (from higher up) to return an empty rectangle if (rc.IsEmpty || xxxButtonSize.Width + 6 > rc.Width). We could either call it only if the width is less than the width of the two buttons plus the minimum required margin or we could have it also take a "normal" left offset to be used when the buttons fit side by side, in which case it would also nicely handle the final calculation for the non-squeezed case.

src/SayMore/Transcription/UI/SegmentingAndRecording/SegmenterDlgBase.cs
```diff
@@ -286,6 +286,11 @@ protected Rectangle GetPlayOrigButtonRectangleForSegment(Rectangle rc)
 			if (rc.IsEmpty || _playButtonSize.Width + 6 > rc.Width)
 				return Rectangle.Empty;
 
+			if (rc.Width < 45)
+				return new Rectangle(rc.X + 3,
```
tombogle, line 290: Similarly, the 3 and 6 here should be constants to ensure everything stays aligned. (Or make the aforementioned helper method smart enough to take a collection of "other" buttons that need to fit on the same row. In this case, there would be no others. Not sure though - that might over-complicate the logic and make it less readable.)

src/SayMore/Transcription/UI/SegmentingAndRecording/OralAnnotationRecorderDlgBase.cs
```diff
@@ -1550,6 +1549,16 @@ private Rectangle GetRerecordButtonRectangleForSegmentMouseIsOver()
 				return Rectangle.Empty;
 			}
 
+			// SP-1000: Reduce minimum segment size
```
tombogle, line 1552: I think you must have intended to call GetButtonRectangleForSegment for this one also?

PR #118, opened 2020-09-30
Buttons not in last segment if segment less than 0.5 seconds

phillip-hopper:
Also fixed floating point errors that started happening after updating Windows and Visual Studio

### DIFF
--- a/src/SayMore/Transcription/Model/TimeTier.cs
+++ b/src/SayMore/Transcription/Model/TimeTier.cs
@@ -296,7 +296,7 @@ namespace SayMore.Transcription.Model
 		/// ------------------------------------------------------------------------------------
 		public bool GetIsAcceptableSegmentLength(float start, float end)
 		{
-			return end - start >= Settings.Default.MinimumSegmentLengthInMilliseconds / 1000f;
+			return Math.Round(end * 1000 - start * 1000) >= Settings.Default.MinimumSegmentLengthInMilliseconds;
 		}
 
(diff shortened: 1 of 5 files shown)


tombogle:
I wonder if this is the desired behavior? I now recall having a discussion with @hatton (years back) and deciding that if the user left a small segment at the end that it was most likely just final cruft in the recording and would not need to be annotated. Now that we've decided that a valid segment can be less than 0.5 seconds, maybe we want to adjust the threshold for this as well, but still leave that logic in place. But maybe that was a bad decision, given Sarah's feedback.

src/SayMore/Transcription/Model/TimeTier.cs
```diff
@@ -296,7 +296,7 @@ public BoundaryModificationResult ChangeSegmentsEndBoundary(AnnotationSegment se
 		/// ------------------------------------------------------------------------------------
 		public bool GetIsAcceptableSegmentLength(float start, float end)
 		{
-			return end - start >= Settings.Default.MinimumSegmentLengthInMilliseconds / 1000f;
+			return Math.Round(end * 1000 - start * 1000) >= Settings.Default.MinimumSegmentLengthInMilliseconds;
```
tombogle, line 299: What sort of error was occurring? If this is the necessary fix, probably a comment is needed, lest it be put back later by someone wishing to make it more readable.

tombogle (changes requested):

PR #128, opened 2021-04-27
SP-2228: Fixed calculation of total duration to not count both the source and "StandardAudio"

tombogle:
Also fixed a little problem (using a bit of a hack) where using F5 or the context menu to refresh the statistics view made it go blank.

### DIFF
--- a/src/SayMore/MediaUtils/MediaFileInfo.cs
+++ b/src/SayMore/MediaUtils/MediaFileInfo.cs
@@ -117,7 +117,7 @@ namespace SayMore.Media
 		/// Gets the total duration in seconds. For audio files, the duration is always the
 		/// duration of the audio. For video files, it is the total duration, counting from the
 		/// start of the first track to the end of the last track (audio and video tracks are
-		/// not guaranteed to start and end simultaneouly).
+		/// not guaranteed to start and end simultaneously).
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		public float DurationInSeconds
--- a/src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs
+++ b/src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.IO;
+using SayMore.Media;
+
+namespace SayMore.Model.Files.DataGathering
+{
+	class SourceAndStandardAudioCoalescingComparer : IEqualityComparer<MediaFileInfo>
+	{
+		private Regex m_regexStandardAudio = new Regex(Properties.Settings.Default.StandardAudioFileSuffix + "$",
+			RegexOptions.Compiled | RegexOptions.IgnoreCase);
+	
+		public bool Equals(MediaFileInfo x, MediaFileInfo y)
+		{
+			if (ReferenceEquals(x, y))
+				return true;
+			if (ReferenceEquals(x, null))
+				return false;
+			if (ReferenceEquals(y, null))
+				return false;
+			if (x.GetType() != y.GetType())
+				return false;
+			if (Math.Abs(x.DurationInMilliseconds - y.DurationInMilliseconds) > 1500) // Somewhat arbitrary fudge factor
+				return false;
+			if (x.MediaFilePath.Equals(y.MediaFilePath))
+				return true;
+			var matchX = m_regexStandardAudio.Match(x.MediaFilePath);
+			if (matchX.Success)
+				return y.MediaFilePath.StartsWith(x.MediaFilePath.Substring(0, matchX.Index));
+			var matchY = m_regexStandardAudio.Match(y.MediaFilePath);
+			return matchY.Success && x.MediaFilePath.StartsWith(y.MediaFilePath.Substring(0, matchY.Index));
+		}
+
+		public int GetHashCode(MediaFileInfo obj)
+		{
+			unchecked
+			{
+				var directory = obj.MediaFilePath == null ? null : Path.GetDirectoryName(obj.MediaFilePath);
+				// We'd prefer to include the DurationInMilliseconds in the hashcode, but unfortunately a
+				// converted file can vary slightly in duration.
+				return directory != null ? directory.GetHashCode() : 0;
+			}
+		}
+	}
+}
--- a/src/SayMore/SayMore.csproj
+++ b/src/SayMore/SayMore.csproj
@@ -241,6 +241,7 @@
     <Compile Include="MediaUtils\Audio\AudioUtils.cs" />
     <Compile Include="MediaUtils\Audio\WaveFileReaderWrapper.cs" />
     <Compile Include="Model\ArchivingHelper.cs" />
+    <Compile Include="Model\Files\DataGathering\SourceAndStandardAudioCoalescingComparer.cs" />
     <Compile Include="Model\IIMDIArchivable.cs" />
     <Compile Include="Model\IRAMPArchivable.cs" />
     <Compile Include="ResourceImageCache.cs" />
@@ -1053,6 +1054,7 @@
     </EmbeddedResource>
     <EmbeddedResource Include="UI\Overview\Statistics\StatisticsView.resx">
       <DependentUpon>StatisticsView.cs</DependentUpon>
+      <SubType>Designer</SubType>
     </EmbeddedResource>
     <EmbeddedResource Include="UI\ProjectWindow\ShowHtmlDialog.resx">
       <DependentUpon>ShowHtmlDialog.cs</DependentUpon>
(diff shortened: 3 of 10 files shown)


hatton (changes requested):
> Reviewed 10 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/128)* status: 10 of 14 files reviewed, 5 unresolved discussions (waiting on @tombogle)

___
*[src/SayMore/Model/ProjectElement.cs, line 502 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUDwzADq_DqQPRXMNt:-MZUDwzADq_DqQPRXMNu:bwdl5th) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/ProjectElement.cs#L502)):*
> ```Smalltalk
> 			var totalTime = new TimeSpan();
> 			var mediaFilesAlreadyProcessed = new HashSet<MediaFileInfo>(new SourceAndStandardAudioCoalescingComparer());
> 			foreach (var file in files)
> ```

I feel dumb, but I'm missing something, can you add two comments to help?

Here we're creating a hashset in order to avoid doing something twice. But unless two or more files can have the same MediaFileInfo, then it's meaningless, since the foreach is already going to do each one only once.

Are some files supposed to not have a MediaFileInfo? 
___
*[src/SayMore/Model/Files/ComponentFile.cs, line 303 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUGyraEW3MWH5Vbang:-MZUGyraEW3MWH5Vbanh:bkhap0) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/ComponentFile.cs#L303)):*
> ```Smalltalk
> 		}
> 
> 		public MediaFileInfo MediaFileInfo => StatisticsProvider?.GetFileData(PathToAnnotatedFile);
> ```

If `GetFileData` isn't super cheap, then I think this would be better as a normal function. When things look like properties, people use them as though they are free. And if in fact they are expensive, then things slow down. E.g., I would not write the code you have above where you extract this value and save it locally and then use it. I would simple using file.MediaFileInfo 3 times, thinking that that is actually easier code to read. 

Next, because c# has this lame characteristic that everything can be null, and in fact GetFileData does intentionally return null sometimes, consider naming this function `GetMedialFileOrNull()`. Please add a public comment here saying when it will be null.
___
*[src/SayMore/Model/Files/ComponentFile.cs, line 313 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUIhqwBnBJqBw91m1a:-MZUIhqwBnBJqBw91m1b:b7lkco8) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/ComponentFile.cs#L313)):*
> ```Smalltalk
> 					return TimeSpan.Zero;
> 
> 				var stats = MediaFileInfo;
> ```

So we have `GetFileData` and `MediaFileInfo`. Can those names be harmonized?  If `GetFileData` actually returns a MediaFIleInfo, then `GetMediaFileInfoOrNull()` would be better there, too.
___
*[src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs, line 9 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUKdo42tWo8h2Kj3GI:-MZUKdo42tWo8h2Kj3GJ:bopzr7r) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs#L9)):*
> ```Smalltalk
> namespace SayMore.Model.Files.DataGathering
> {
> 	class SourceAndStandardAudioCoalescingComparer : IEqualityComparer<MediaFileInfo>
> ```

What would be a good class comment for this? I don't completely understand, so here's my guess:

``
From a single original artifact, we may create multiple derivative files. E.g. both a video and an audio-only extract. When we go to do statistics on the project, we only want to  count the duration of the original file. But it's easier (for some reason?) to just compare all the files and drop the ones that appear to be derivatives
``
___
*[src/SayMore/UI/Overview/Statistics/StatisticsView.cs, line 75 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUK7CT06m1z2GxxAMd:-MZUK7CT06m1z2GxxAMe:b3vo1i6) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/UI/Overview/Statistics/StatisticsView.cs#L75)):*
> ```Smalltalk
> 							_model.NewStatisticsAvailable += HandleNewDataAvailable;
> 
> 							_detectContextMenuRefreshTimer.Enabled = true;
> ```

I think this is more like `_detectBorkedBrowserTimer`otherwise it looks like it is something to enable context menus.

tombogle:
___
*[src/SayMore/Model/ProjectElement.cs, line 502 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUDwzADq_DqQPRXMNt:-MZUSx4gE96HafPxc7r-:b-cc1c6z) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/ProjectElement.cs#L502)):*

I feel dumb, but I'm missing something, can you add two comments to help?

Here we're creating a hashset in order to avoid doing something twice. But unless two or more files can have the same MediaFileInfo, then it's meaningless, since the foreach is already going to do each one only once.

Are some files supposed to not have a MediaFileInfo? 


</blockquote></details>

I have added a comment that should clarify the (admittedly no terribly intuitive) intention of this code. This is going through all the files except the generated audio file, so it will include metadata files, etc.  that do not have MediaFileInfo.

tombogle:
___
*[src/SayMore/Model/Files/ComponentFile.cs, line 303 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUGyraEW3MWH5Vbang:-MZUVJr578qJ_RsklNtV:bd7zwfl) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/ComponentFile.cs#L303)):*

If `GetFileData` isn't super cheap, then I think this would be better as a normal function. When things look like properties, people use them as though they are free. And if in fact they are expensive, then things slow down. E.g., I would not write the code you have above where you extract this value and save it locally and then use it. I would simple using file.MediaFileInfo 3 times, thinking that that is actually easier code to read. 

Next, because c# has this lame characteristic that everything can be null, and in fact GetFileData does intentionally return null sometimes, consider naming this function `GetMedialFileOrNull()`. Please add a public comment here saying when it will be null.
</blockquote></details>

Fair point, though not a principle widely observed in SayMore. The following property DurationSeconds actually accesses the MediaFileInfo property, so it's even worse, and DurationString accesses DurationSeconds. But I made your suggested change and even refactored so that when projectElement gets it, it can re-use the media info it gets to compute the Duration in seconds.

tombogle:
___
*[src/SayMore/Model/Files/ComponentFile.cs, line 313 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUIhqwBnBJqBw91m1a:-MZUXYfN5F9A-iQyeLnB:b-gnfyfo) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/ComponentFile.cs#L313)):*
> GetMediaFileInfoOrNull
The method on the StatisticsProvider is named more generically (GetFileData) because it is in an interface with a generic return type. I've renamed the specific one (whose return type is known to be MediaFileInfo) to GetMediaFileInfoOrNull as requested. There is currently only one other method in SayMore that uses the "OrNull" naming convention. Of course, in a perfect world, we'd move up to a version of c# that would allow methods to be explicit about whether they can return null.

tombogle:
___
*[src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs, line 9 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUKdo42tWo8h2Kj3GI:-MZUa9S_8YF9Op956csH:b-5n6nd5) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/Model/Files/DataGathering/SourceAndStandardAudioCoalescingComparer.cs#L9)):*

What would be a good class comment for this? I don't completely understand, so here's my guess:

``
From a single original artifact, we may create multiple derivative files. E.g. both a video and an audio-only extract. When we go to do statistics on the project, we only want to  count the duration of the original file. But it's easier (for some reason?) to just compare all the files and drop the ones that appear to be derivatives
``
</blockquote></details>

That was a good start. I made it three times longer :-)

tombogle:
___
*[src/SayMore/UI/Overview/Statistics/StatisticsView.cs, line 75 at r1](https://reviewable.io/reviews/sillsdev/saymore/128#-MZUK7CT06m1z2GxxAMd:-MZUalUR4hgP_TRJ0tBR:bv0t1fp) ([raw file](https://github.com/sillsdev/saymore/blob/2a6e13067fe9797075c282f2820f77a5129c714a/src/SayMore/UI/Overview/Statistics/StatisticsView.cs#L75)):*
> ```
> _detectContextMenuRefreshTimer
> ```
How about _timerDetectBrowserRefreshedUsingContextMenu? (I'd hate to lose the information about what caused the browser to get borked unless/until we find that there are more ways to make it happen.)

PR #134, opened 2021-06-04
SP-2197: Fixed crash when array with samples does not get filled

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/134)


tombogle:
@phillip-hopper , I worked on this a few weeks ago and then left it as a draft. Now that I'm coming back to it, I can't see what else I might have hoped to do, without actually being able to get a file that makes the crash reproducible. If you see something that looks unfinished, let me know. Otherwise, hopefully, this avoids the actual crash and gives us more information if this situation ever arises again.

phillip-hopper (approved):
LGTM :+1:

PR #135, opened 2021-07-13
SP-2333: Prevented undesirable behavior related to MS Office temp files

tombogle:
Excluded files that appear to be MS Office temp files from Project docs list. Also, made file list a read-only collection so that underlying file system changes could not occur undetected, leading to crashes when working with the ComponentFile grid.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/135)


phillip-hopper (approved):
LGTM :+1:

PR #136, opened 2021-11-09
SP-2198: Handle problems in font settings string stored in sprj file

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/136)

### DIFF
--- a/src/SayMore/Model/Project.cs
+++ b/src/SayMore/Model/Project.cs
@@ -24,7 +24,7 @@ using SIL.Archiving.IMDI;
 using SayMore.Properties;
 using SayMore.Transcription.Model;
 using SayMore.Model.Files;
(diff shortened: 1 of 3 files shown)


phillip-hopper:
@tombogle, Do you have an example of a font setting that will trigger the fallback code?

tombogle:
I was unable to find a way inside SayMore to recreate the scenario, so I'm not sure how it got this way. I just manually edited one of my sprj files to set the size to 0. (For completeness, also did the same thing to try it with a negative number. I also tried a really tiny decimal number, which seemed to work fine, although it resulted in a font that was too tiny to be legible.)

phillip-hopper (approved):
LGTM :+1:

PR #145, opened 2022-05-11
SP-1760: Added defensive code to prevent argument out of range error

tombogle:
Although I was not able to reproduce this error (it is several years old, and possibly some other fix now makes it impossible), it seemed best to check the index rather than blindly accessing a potentially nonexistent element merely for the sake of being able to re-select it later.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/145)


tombogle:
I think found the

tombogle:
I think I found the actual fix (made for SP-2233 and released in v. 3.2.21). I think this might still be a good check to do, but I'm kind of on the fence, since it does not appear to be reproducible.

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/145)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

PR #146, opened 2022-05-16
SP-2269: If "other" component file fails to load, allow project to load.

tombogle:
If it is an apparent temp file, just log it and don't even bother the user.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/146)


andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/146)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/SayMore/Model/ProjectElement.cs` line 180 at r1](https://reviewable.io/reviews/sillsdev/saymore/146#-N2DCOc8-gkcD5uLEJQ7:-N2DCOc8-gkcD5uLEJQ8:b-9ow0w7) ([raw file](https://github.com/sillsdev/saymore/blob/ea047102ee4c289c3e503ea8cb32dc8210130142/src/SayMore/Model/ProjectElement.cs#L180)):*
> ```Smalltalk
> 									"Failed to load file:\r{0}\rIf this is not an essential" +
> 									" file, you can safely delete it and/or ignore this error."),
> 								filename);
> ```

Is it correct/normal to do line breaks with just `\r`?

Is it worth doing anything to prevent the user from getting a bunch of these messages one after the other?

tombogle:
*[`src/SayMore/Model/ProjectElement.cs` line 180 at r1](https://reviewable.io/reviews/sillsdev/saymore/146#-N2DCOc8-gkcD5uLEJQ7:-N2DaHtDD8HQE7jku-7z:b-ai777b) ([raw file](https://github.com/sillsdev/saymore/blob/ea047102ee4c289c3e503ea8cb32dc8210130142/src/SayMore/Model/ProjectElement.cs#L180)):*

Is it correct/normal to do line breaks with just `\r`?

Is it worth doing anything to prevent the user from getting a bunch of these messages one after the other?
</blockquote></details>

I originally put in \r\n, but Resharper begged me to remove the \n. I just checked it, and both seem to be required. Good thing I checked it because if SayMore takes this path through the code and displays the NotifyUser dialog and the user ignores it (which is the normal thing to do), then SayMore crashes on startup! More work needed...
These messages should be very rare! In the unlikely case that there is more than once "corrupt" file in the project that does not appear to be a temp file, presumably the user would need to be alerted to each one. I suppose there could be an odd case of massive corruption, but then it is likely that it would affect the essential files as well. Perhaps in a worst-case scenario, we'd want to make the alternate button I think I feel

andrew-polk (changes requested):
> Reviewed 63 of 63 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/146)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/SayMore/Model/Session.cs` line 270 at r2](https://reviewable.io/reviews/sillsdev/saymore/146#-N2JKPKI8YBiW1gHcck8:-N2JKPKI8YBiW1gHcck9:buvsner) ([raw file](https://github.com/sillsdev/saymore/blob/462ca6614f31f5c0bab94024d444248dcbdafd88/src/SayMore/Model/Session.cs#L270)):*
> ```Smalltalk
> 		private void ProcessContributorNameChange(ElementIdChangedArgs e)
> 		{
> 			foreach (var file in GetComponentFiles().Where(f => f.FileType is FileTypeWithContributors))
> ```

Nothing wrong with this, but you can use `OfType` instead if you want.
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.oftype?view=net-6.0
___
*[`src/SayMore/UI/ElementListScreen/ComponentFileGrid.cs` line 100 at r2](https://reviewable.io/reviews/sillsdev/saymore/146#-N2JMb2T9OEG09nxf7Sz:-N2JMb2T9OEG09nxf7T-:b-mdj8ko) ([raw file](https://github.com/sillsdev/saymore/blob/462ca6614f31f5c0bab94024d444248dcbdafd88/src/SayMore/UI/ElementListScreen/ComponentFileGrid.cs#L100)):*
> ```Smalltalk
> 		protected virtual void HandleStringsLocalized(ILocalizationManager lm)
> 		{
> 			if (_grid == null || _grid.IsDisposed || lm.Id != ApplicationContainer.kSayMoreLocalizationId)
> ```

Not sure what the implications are, but everywhere else, you check for null `lm`
___
*[`src/SayMore/UI/Overview/Statistics/StatisticsView.cs` line 56 at r2](https://reviewable.io/reviews/sillsdev/saymore/146#-N2JNU57BciC2RqZ58_q:-N2JNU57BciC2RqZ58_r:b-9q54b5) ([raw file](https://github.com/sillsdev/saymore/blob/462ca6614f31f5c0bab94024d444248dcbdafd88/src/SayMore/UI/Overview/Statistics/StatisticsView.cs#L56)):*
> ```Smalltalk
> 
> 		/// ------------------------------------------------------------------------------------
> 		private void UpdateDisplay(ILocalizationManager lm = null)
> ```

not used



_Code quote:_
```Smalltalk
ILocalizationManager lm = null
```

tombogle:
*[`src/SayMore/UI/Overview/Statistics/StatisticsView.cs` line 56 at r2](https://reviewable.io/reviews/sillsdev/saymore/146#-N2JNU57BciC2RqZ58_q:-N2MPVGlAQYt2aWPQO6Z:bnvfm86) ([raw file](https://github.com/sillsdev/saymore/blob/462ca6614f31f5c0bab94024d444248dcbdafd88/src/SayMore/UI/Overview/Statistics/StatisticsView.cs#L56)):*

not used
</blockquote></details>

I wasn't 100% sure that we wouldn't need to refresh it in the case of Palaso strings changing, but now I've researched it more deeply and we don't. That said, I'm having a hard time understanding the plumbing in L10nSharp because this only seems to be getting called for the SayMore LM, never for the Palaso one, even though all the strings are presented there. Research continues....

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/146)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`src/SayMore/UI/ElementListScreen/ComponentFileGrid.cs` line 100 at r2](https://reviewable.io/reviews/sillsdev/saymore/146#-N2JMb2T9OEG09nxf7Sz:-N2MP1lm1tE6FbpcVfEi:b-fe0mbr) ([raw file](https://github.com/sillsdev/saymore/blob/462ca6614f31f5c0bab94024d444248dcbdafd88/src/SayMore/UI/ElementListScreen/ComponentFileGrid.cs#L100)):*

Not sure what the implications are, but everywhere else, you check for null `lm`
</blockquote></details>

Not necessary here. I have tweaked the code to make that more obvious.

tombogle:
Waiting on changes from https://github.com/sillsdev/l10nsharp/pull/97

PR #152, opened 2022-08-01
Switched to SIL.DesktopAnalytics and fixed DLL issues with Newtonsoft.json

tombogle:
Upgraded lots of DLLs

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/152)

### DIFF
--- a/build/SayMore.proj
+++ b/build/SayMore.proj
@@ -3,7 +3,7 @@
 	<RootDir Condition="'$(teamcity_version)' == ''">$(MSBuildProjectDirectory)\..</RootDir>
 	<RootDir Condition="'$(teamcity_version)' != ''">$(teamcity_build_checkoutDir)</RootDir>
 	<BUILD_NUMBER Condition="'$(BUILD_NUMBER)'==''">3.3.0</BUILD_NUMBER>
-	<BuildTasksDll>$(RootDir)/packages/SIL.BuildTasks.2.2.0/tools/SIL.BuildTasks.dll</BuildTasksDll>
+	<BuildTasksDll>$(RootDir)/packages/SIL.BuildTasks.2.5.0/tools/SIL.BuildTasks.dll</BuildTasksDll>
 	<Configuration>Release</Configuration>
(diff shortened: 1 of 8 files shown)


andrew-polk (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/152)* status: all files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor and @tombogle)

___
*[`src/SayMoreTests/app.config` line 23 at r1](https://reviewable.io/reviews/sillsdev/saymore/152#-N8_k-AjA8YzKvxbzmIy:-N8_k-Ak-MXz5CzVFpQE:b1p76rz) ([raw file](https://github.com/sillsdev/saymore/blob/c0346f1bb5e0acc3b1765f267d1e1a7dd6fa256c/src/SayMoreTests/app.config#L23)):*
> ```
>         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
>         <bindingRedirect oldVersion="0.0.0.0-12.0.0.0" newVersion="12.0.0.0" />
>       </dependentAssembly>
> ```

I'm surprised to see this removed but don't have any specific reason to believe it will cause a problem. Presumably this was an automatic change it made because it noticed all the packages which rely on it are using the same version?

Otherwise, LGTM

tombogle:
*[`src/SayMoreTests/app.config` line 23 at r1](https://reviewable.io/reviews/sillsdev/saymore/152#-N8_k-AjA8YzKvxbzmIy:-N8_mPb-10OQ78IDiq15:bi7p1hx) ([raw file](https://github.com/sillsdev/saymore/blob/c0346f1bb5e0acc3b1765f267d1e1a7dd6fa256c/src/SayMoreTests/app.config#L23)):*

I'm surprised to see this removed but don't have any specific reason to believe it will cause a problem. Presumably this was an automatic change it made because it noticed all the packages which rely on it are using the same version?

Otherwise, LGTM
</blockquote></details>

At least in theory now, everything should have moved onto the latest version because v. 12 had some kind of security vulnerability.

PR #156, opened 2022-09-07
SP-1769: Prevent changing the grid while executing PlaybackProgressReportAction

tombogle:
Also check for a null grid before accessing it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/156)


tombogle:
I have not been able to reproduce this, but I'm guessing this will fix the crash. (Of course, that's what I thought the last time I "fixed" it. I'm hoping to get the current user to give me more details to actually be able to repro.

ermshiperete (approved):

PR #168, opened 2023-01-04
Report condition where system volume is not set up to compute short names

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/168)

### DIFF
--- a/SayMore.sln.DotSettings
+++ b/SayMore.sln.DotSettings
@@ -8,6 +8,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=bldr/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=curr/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Dropbox/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=fsutil/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Insite/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Mets/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Palaso/@EntryIndexedValue">True</s:Boolean>
--- a/src/SayMore/MediaUtils/MPlayer/MPlayerHelper.cs
+++ b/src/SayMore/MediaUtils/MPlayer/MPlayerHelper.cs
@@ -147,7 +147,9 @@ namespace SayMore.Media.MPlayer
 			output = null;
 
 			// problem with non-ascii characters, convert to 8.3 path
-			mediaInPath = FileSystemUtils.GetShortName(mediaInPath);
+			mediaInPath = FileSystemUtils.GetShortName(mediaInPath,
+                () => LocalizationManager.GetString("CommonToMultipleViews.MediaPlayer.ProbablyCannotLoad",
+                    "SayMore will probably not be able to convert this file."));
 
 			// output to temp file, with 8.3 path
 			var tempFile = Path.GetTempFileName();
--- a/src/SayMore/MediaUtils/MPlayer/MediaPlayer.cs
+++ b/src/SayMore/MediaUtils/MPlayer/MediaPlayer.cs
@@ -65,6 +65,10 @@ namespace SayMore.Media.MPlayer
             {
                 Logger.WriteError(e);
             }
+            catch (ObjectDisposedException e)
+            {
+                Logger.WriteError(e);
+            }
         }
 
         /// ------------------------------------------------------------------------------------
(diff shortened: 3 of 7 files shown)


andrew-polk (changes requested):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/168)* status: all files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 173 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNU75ABYJ83F6-9FpB:-NLNU75ABYJ83F6-9FpC:b49crrc) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L173)):*
> ```Smalltalk
>             if (result == path && getDescriptionOfFailedAction != null &&
>                 (Path.GetFileName(result).Length > 12 || result.Any(c => c >= 0x80 || c == ' ') ||
>                     result.Count(c => c == '.') > 1))
> ```

Why include the `.Count(c => c == '.') > 1` in the `.Any()`?

`c >= 0x80` probably deserves a comment. Or at least the whole if condition.

Do you really only want to catch and report something here if getDescriptionOfFailedAction is defined?
___
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 196 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNV1S4BSk29WW_tteO:-NLNV1S4BSk29WW_tteP:b-rr646g) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L196)):*
> ```Smalltalk
>                     "file names for the file system volume ({3}) where this file is located.",
>                     "Param 0: \"SayMore\" (product name); " +
>                     "Param 1: ; " +
> ```

forgot to add path to param comment.



_Code quote:_
```Smalltalk
"Param 1: ; " +
```

tombogle:
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 173 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNU75ABYJ83F6-9FpB:-NLU-3bG4LdRQiFohCpl:bdvh6gc) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L173)):*

Why include the `.Count(c => c == '.') > 1` in the `.Any()`?

`c >= 0x80` probably deserves a comment. Or at least the whole if condition.

Do you really only want to catch and report something here if getDescriptionOfFailedAction is defined?
</blockquote></details>

An 8.3 filename is only allowed to have a single period, so I'm trying to do a reasonable job of making sure we got a valid short name. Thinking about it, probably a regex would be better. There's one (or maybe two?) case where this can get called when the user is not doing something specific that makes reporting the problem meaningful. Some information is just being displayed in a browser control I think, so I figured it was better to just wait until they tried to do something where it would really matter.

tombogle:
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 173 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNU75ABYJ83F6-9FpB:-NLUEIUH5KrhIPOOQDm0:bcr0a5t) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L173)):*

An 8.3 filename is only allowed to have a single period, so I'm trying to do a reasonable job of making sure we got a valid short name. Thinking about it, probably a regex would be better. There's one (or maybe two?) case where this can get called when the user is not doing something specific that makes reporting the problem meaningful. Some information is just being displayed in a browser control I think, so I figured it was better to just wait until they tried to do something where it would really matter. 
</blockquote></details>

I made the change to use a regex. Thought I'd find one ready-made on the web, but I found a lot of half-hearted attempts instead.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/168)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @andrew-polk and @tombogle)

___
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 196 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNV1S4BSk29WW_tteO:-NLUEI0C30nOR0oibdav:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L196)):*

forgot to add path to param comment.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/168)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`src/SayMore/Utilities/FileSystemUtils.cs` line 173 at r1](https://reviewable.io/reviews/sillsdev/saymore/168#-NLNU75ABYJ83F6-9FpB:-NLX8cHb2MAsdR3R1I35:b1zgwhw) ([raw file](https://github.com/sillsdev/saymore/blob/c447a91c4db284be21375a4662d526d160991de4/src/SayMore/Utilities/FileSystemUtils.cs#L173)):*

I made the change to use a regex. Thought I'd find one ready-made on the web, but I found a lot of half-hearted attempts instead.
</blockquote></details>

Sorry; that was supposed to say, "Why not include..." and I was only suggesting a simplification. But the regex is probably better.

PR #176, opened 2023-05-15
SP-2307: prevent Invoke before MediaPlayer window handle is created

tombogle:
Changed sequence of setup operations in MediaPlayer to prevent setting handlers that might be called and Invoke an operation before the window handle is created.
Note: I have not found a way to reproduce this crash, but it was reported twice by the same user in quick succession.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/176)


andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/176)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

PR #179, opened 2023-09-06
Added ability to set the working language font

tombogle:
https://community.software.sil.org/t/feature-request-ability-to-change-font-size-in-note-section/4401

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/179)


andrew-polk (changes requested):
> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/179)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/SayMore/Model/Project.cs` line 261 at r1](https://reviewable.io/reviews/sillsdev/saymore/179#-Ndphnbp2iQFZYZmA44Q:-Ndphnbp2iQFZYZmA44R:b38aybo) ([raw file](https://github.com/sillsdev/saymore/blob/b2af54317d6a39bb9fefd307def9ce2bd37072ac/src/SayMore/Model/Project.cs#L261)):*
> ```Smalltalk
> 					FontHelper.FontToString(_workingLanguageFont)));
> 
> 			var autoSegmenterSettings = new XElement(kAutoSegmenterSettings);
> ```

Doesn't this create a backward compatibility issue?
I.e. 3.5 won't be able to use settings saved in 3.6?

tombogle:
*[`src/SayMore/Model/Project.cs` line 261 at r1](https://reviewable.io/reviews/sillsdev/saymore/179#-Ndphnbp2iQFZYZmA44Q:-NdqhX5uBwwE5gRvEx7b:b3tuidi) ([raw file](https://github.com/sillsdev/saymore/blob/b2af54317d6a39bb9fefd307def9ce2bd37072ac/src/SayMore/Model/Project.cs#L261)):*

Doesn't this create a backward compatibility issue?
I.e. 3.5 won't be able to use settings saved in 3.6?
</blockquote></details>

True. Sad! I guess I'll have to revert the spelling/capitalization changes...

tombogle:
*[`src/SayMore/Model/Project.cs` line 261 at r1](https://reviewable.io/reviews/sillsdev/saymore/179#-Ndphnbp2iQFZYZmA44Q:-Ne3aKtj31RFb-ETnc85:b-bjn1oh) ([raw file](https://github.com/sillsdev/saymore/blob/b2af54317d6a39bb9fefd307def9ce2bd37072ac/src/SayMore/Model/Project.cs#L261)):*

True. Sad! I guess I'll have to revert the spelling/capitalization changes...
</blockquote></details>

Upon looking at the code further, I see that there is no way to change these settings in the program. So the only way they would ever be different from the defaults would be if someone changed them manually. While this is certainly possible, I sort of wonder if the risk is really high enough to worry about. Also, considering that the settings would have to be modified manually, that's all the more reason to spell them correctly and make the setting easier to find and reduce the chances that an over-zealous user would fix the spelling error and then have it be ignored. What do you think?

andrew-polk (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/179)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`src/SayMore/Model/Project.cs` line 261 at r1](https://reviewable.io/reviews/sillsdev/saymore/179#-Ndphnbp2iQFZYZmA44Q:-Ne4Gv0VCZZOlIh3h1jK:bp2wpcu) ([raw file](https://github.com/sillsdev/saymore/blob/b2af54317d6a39bb9fefd307def9ce2bd37072ac/src/SayMore/Model/Project.cs#L261)):*

Upon looking at the code further, I see that there is no way to change these settings in the program. So the only way they would ever be different from the defaults would be if someone changed them manually. While this is certainly possible, I sort of wonder if the risk is really high enough to worry about. Also, considering that the settings would have to be modified manually, that's all the more reason to spell them correctly and make the setting easier to find and reduce the chances that an over-zealous user would fix the spelling error and then have it be ignored. What do you think?
</blockquote></details>

If we're only talking about manual modification, it doesn't seem worth worrying about. Unless this feature was specifically created so someone could manually modify it... (I'd be curious why we even bother persisting them, otherwise...)

tombogle:
*[`src/SayMore/Model/Project.cs` line 261 at r1](https://reviewable.io/reviews/sillsdev/saymore/179#-Ndphnbp2iQFZYZmA44Q:-Ne4I3JN1CT9LZ72B3wQ:beiy4xx) ([raw file](https://github.com/sillsdev/saymore/blob/b2af54317d6a39bb9fefd307def9ce2bd37072ac/src/SayMore/Model/Project.cs#L261)):*

If we're only talking about manual modification, it doesn't seem worth worrying about. Unless this feature was specifically created so someone could manually modify it... (I'd be curious why we even bother persisting them, otherwise...)
</blockquote></details>

I don't specifically recall (though I'm sure I coded it), but I suspect we figured there was a slight chance that someone would need to modify the defaults, so we wanted to make it possible. Maybe we thought we'd come back and add UI later.

PR #180, opened 2023-12-11
Upgrade desktop analytics

tombogle:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/180)


andrew-polk (approved):
I'm a bit surprised to see the mixpanel dep come in, but if that seems right to you, go ahead.

PR #217, opened 2025-04-15
Solve undesirable behavior and deadlocks in text annotation grid

tombogle:
SP-2306: This most likely solves problem of the first cell getting focus unexpectedly on Mac running via Parallels.
It also solves a deadlock when right-clicking to display the context menu.
Probably generally avoids other potential deadlock situations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/217)

### DIFF
--- a/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs
+++ b/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs
@@ -15,6 +15,7 @@ using SayMore.Model.Files;
 using SayMore.Properties;
 using SayMore.Transcription.Model;
 using SayMore.Media.MPlayer;
+using static SIL.Windows.Forms.Extensions.ControlExtensions.ErrorHandlingAction;
 using GridSettings = SIL.Windows.Forms.Widgets.BetterGrid.GridSettings;
 
 namespace SayMore.Transcription.UI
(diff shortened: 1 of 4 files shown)


src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs
```diff
@@ -407,21 +422,22 @@ protected override void OnCellMouseClick(DataGridViewCellMouseEventArgs e)
...
 			var menuItems = col.GetContextMenuCommands().ToArray();
 			if (menuItems.Length == 0)
 				return;
 
-			CurrentCell = Rows[e.RowIndex].Cells[e.ColumnIndex];
+			var clickedCell = Rows[e.RowIndex].Cells[e.ColumnIndex];
+			// To avoid possibility of churn or weird side effects (esp. on Mac Parallels),
+			// let's only set this if the click occured in a cell other than the current one.
+			if (CurrentCell != clickedCell)
+				CurrentCell = clickedCell;
```
andrew-polk, line 437: You don't need to use InvokeSetCurrentCell here?
If not using it, probably worth a comment?

andrew-polk (changes requested):

src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs
tombogle: No, because OnCellMouseClick can only ever be called on the UI thread. (I had the same doubt, but ChatGPT assured me this was correct.) I can add a comment to that effect.

andrew-polk (approved):

PR #225, opened 2025-07-01
SP-2294: Allow manual addition of session files that the user created with SayMore running

tombogle:


tombogle:
I have broken SP-2294 into two subtasks because this part (SP-2377) was relatively easy, but the other part is trickier.

PR #230, opened 2025-07-07
SP-2371: Added try-catch to avoid reporting reentrant call to SetCurrentCellAddressCore

tombogle:
Also, added early return to RefreshComponentFiles when disposed/disposing (though probably not the cause) Did a little bit of semi-related code cleanup & fixed some spelling errors

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/230)

### DIFF
--- a/SayMore.sln.DotSettings
+++ b/SayMore.sln.DotSettings
@@ -46,5 +46,6 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=toolstrip/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Uncolonized/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Unignored/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=unnannotated/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=unrounded/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Wikimedia/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
--- a/src/SayMore/MediaUtils/MPlayer/MediaPlayer.cs
+++ b/src/SayMore/MediaUtils/MPlayer/MediaPlayer.cs
@@ -7,6 +7,7 @@ using System.Windows.Forms;
 using SIL.Reporting;
 using SayMore.UI.LowLevelControls;
 using SIL.Windows.Forms;
+using SIL.Windows.Forms.Extensions;
 
 namespace SayMore.Media.MPlayer
 {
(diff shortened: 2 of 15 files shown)


andrew-polk (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/230)* status: all files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs` line 659 at r3](https://reviewable.io/reviews/sillsdev/saymore/230#-OUfKpP_1YHqIVdyBwGH:-OUfKpP_1YHqIVdyBwGI:b-upj66j) ([raw file](https://github.com/sillsdev/saymore/blob/9e1e2d5c89f9c2820d996ad7f0f24c47210108c6/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs#L659)):*
> ```Smalltalk
> 		private bool InternalPlay()
> 		{
> 			bool loaded = false;
> ```

Would a better name be something like loadedAtLeastOneFile?
Or anyLoaded?
___
*[`src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs` line 686 at r3](https://reviewable.io/reviews/sillsdev/saymore/230#-OUfKXcM6kjmJ5ylWnBj:-OUfKXcN1lTdHXBliRRX:b-obmbke) ([raw file](https://github.com/sillsdev/saymore/blob/9e1e2d5c89f9c2820d996ad7f0f24c47210108c6/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs#L686)):*
> ```Smalltalk
> 			// possible deadlock, we release the lock before even attempting to invoke
> 			// notifying the user of any problems.
> 			foreach (var errorMessage in errorMessages)
> ```

I'm sure you already thought of this. And maybe there are never more than a handful, but just pointing out that if there are lots of files with errors, this would essentially lock up the app from the user's perspective

andrew-polk (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/230)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tombogle)

___
*[`src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs` line 659 at r3](https://reviewable.io/reviews/sillsdev/saymore/230#-OUfKpP_1YHqIVdyBwGH:-OUlDQV_-u8nIt0E8EgT:b-4ltnmn) ([raw file](https://github.com/sillsdev/saymore/blob/9e1e2d5c89f9c2820d996ad7f0f24c47210108c6/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs#L659)):*

We ever only load one file. (Just might not be the one we were originally hoping to load.) I think the error case is rare enough so as not to be worth confusing things with a longer name.
</blockquote></details>

See egg on my face comment below.
___
*[`src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs` line 686 at r3](https://reviewable.io/reviews/sillsdev/saymore/230#-OUfKXcM6kjmJ5ylWnBj:-OUlDJX_B5Q84Y5Bs8IN:b-t0ryoc) ([raw file](https://github.com/sillsdev/saymore/blob/9e1e2d5c89f9c2820d996ad7f0f24c47210108c6/src/SayMore/Transcription/UI/TextAnnotationGrid/TextAnnotationEditorGrid.cs#L686)):*

Currently never more than 2. The chances of one having an error is super slim. The chances of both having an error is so small I almost didn't bother changing this code to handle that scenario gracefully.
</blockquote></details>

Yeah... totally missed that the loop stops after the first successful load.
I'm ashamed.

PR #236, opened 2025-07-22
Added a label with instructions about importing flextext file into Flex

tombogle:
Though not a "fix" for SP-2378, that issue inspired this improvement. Made a small tweak to the logic to select the correct writing system for translations when the user has set the analysis WS in SayMore. Also made some internationalization/localization improvements in this area of the code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/236)

### DIFF
--- a/DistFiles/SayMore.es.xlf
+++ b/DistFiles/SayMore.es.xlf
@@ -2772,6 +2772,11 @@
         <target xml:lang="es-ES" state="translated">Habla cuidadosa</target>
         <note>ID: DialogBoxes.Transcription.CarefulSpeechRecorderDlg._labelCarefulSpeech</note>
       </trans-unit>
+      <trans-unit id="DialogBoxes.Transcription.CommonAnnotationSegmenterDlg.NoVideoHelpMsg">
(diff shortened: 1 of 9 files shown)


tombogle:
*[`src/SayMore/Transcription/UI/ExportToFieldWorksInterlinearDlg.Designer.cs` line 198 at r1](https://reviewable.io/reviews/sillsdev/saymore/236#-OWMPn-u9mtwjlt9Rmbn:-OWMkC6s0X2YDUfiJ1yr:b9wmf13) ([raw file](https://github.com/sillsdev/saymore/blob/fe655da77e77d10301bf8484748791064e01461b/src/SayMore/Transcription/UI/ExportToFieldWorksInterlinearDlg.Designer.cs#L198)):*

I guess it doesn't matter because you're handling the localization directly, but this is set to `NotLocalizable`.
</blockquote></details>

It matters. In HandleStringsLocalized, I save off the localized text as my format string to use in FormatImportInstructions. Good catch! (There's something twitchy in Designer when using the loc extender. Seems when you "reset" the ID, it also sets it to NotLocalizable. I almost always remember to check that.

PR #243, opened 2025-10-30
Add GitHub Actions workflow for Windows builds

megahirt:
## Summary
- add a Windows GitHub Actions workflow that does build and test on every PR and every commit to master
- creates an installer for every commit on master and signs it
- creates a release for every tag starting with "v"
- allow manual workflow runs
- removes reference to "sign.cmd" and uses codesign instead

## Testing
More testing is needed.  This PR only tests the "build and test" portion, once it's merged. we can manually test the installer creation and signing part.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/saymore/243)

### DIFF
--- a/build/SayMore.proj
+++ b/build/SayMore.proj
@@ -1,9 +1,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
 	<RootDir>$(MSBuildProjectDirectory)\..</RootDir>
-	<RootDir Condition="'$(teamcity_version)' != ''">$(teamcity_build_checkoutDir)</RootDir>
 	<RootDir Condition="'$(GITHUB_WORKSPACE)' != ''">$(GITHUB_WORKSPACE)</RootDir>
-	<BUILD_NUMBER Condition="'$(BUILD_NUMBER)'==''">3.7.100</BUILD_NUMBER>
 	<BuildTasksDll Condition="Exists('$(RootDir)/packages/SIL.BuildTasks.3.1.1/tools/SIL.BuildTasks.dll')">$(RootDir)/packages/SIL.BuildTasks.3.1.1/tools/SIL.BuildTasks.dll</BuildTasksDll>
 	<BuildTasksDll Condition="!Exists('$(RootDir)/packages/SIL.BuildTasks.3.1.1/tools/SIL.BuildTasks.dll')">$(RootDir)/packages/SIL.BuildTasks/tools/SIL.BuildTasks.dll</BuildTasksDll>
 	<SILReleaseTasksProps>$(RootDir)/packages/SIL.ReleaseTasks.3.1.1/build/SIL.ReleaseTasks.props</SILReleaseTasksProps>
@@ -21,7 +19,6 @@
     <Exec Command='$(NuGetCommand) install SIL.BuildTasks -excludeVersion -PreRelease -source "$(PackageSources)" -solutionDirectory "$(SolutionDir)."' />
 	<Exec Command='$(NuGetCommand) install SIL.ReleaseTasks -excludeVersion -PreRelease -source "$(PackageSources)" -solutionDirectory "$(SolutionDir)"' />
 	<Exec Command='$(NuGetCommand) install SIL.libpalaso.l10ns -excludeVersion -PreRelease -source "$(PackageSources)" -solutionDirectory "$(SolutionDir)."' />
-	<Exec Command='$(NuGetCommand) install NUnit.Extension.TeamCityEventListener -excludeVersion -source "$(PackageSources)" -solutionDirectory "$(SolutionDir)."' />
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="BuildInternal"
       Properties="Configuration=$(Configuration)" Condition="$(RestartBuild)" />
   </Target>
@@ -34,6 +31,10 @@
   <UsingTask TaskName="NUnit3" AssemblyFile="$(BuildTasksDll)" Condition="Exists('$(BuildTasksDll)')" />
 
   <Target Name="VersionNumbers">
+	<PropertyGroup>
+		<!-- The Version should be set from the outside, if not, set a default value -->
+		<Version Condition="'$(Version)' == ''">$([System.DateTime]::Now.ToString("yyyyMMddHHmmss"))</Version>
+	</PropertyGroup>
 
 		<Message Text="Version: $(Version)" Importance="high"/>
   </Target>
@@ -73,7 +74,7 @@
 	<ItemGroup>
 	  <TestAssemblies Include="$(RootDir)/output/release/*Tests.dll;"/>
 	</ItemGroup>
-	<NUnit3 Condition="'$(teamcity_version)' == ''"
+	<NUnit3
 		Assemblies="@(TestAssemblies)"
 		ToolPath="$(RootDir)/packages/NUnit.ConsoleRunner.3.20.1/tools"
 		ExcludeCategory="$(excludedCategories)"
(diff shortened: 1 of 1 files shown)


tombogle (changes requested):
I don't see anything in this YAML that deals with the ConvertReleaseNotesToHtml target. The results of that target are only used in an actual publish build, so maybe your thought is that that would be done in a separate GHA. But it is an artifact of the other builds on TC. For the most part, nothing should change that could break this, but if a PR upgraded the BuildTasks DLL, for example, it would be nice to be able to verify that the conversion still worked. 

> @tombogle reviewed 1 of 3 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/243#-:-OcqexNl-EcWEnC_uv4O:b-18hvlh)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @Copilot)

___
*[`.github/workflows/build-test-installer-release.yml` line 45 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-OcqjYxIAA8qL3GFoHy_:-OcqjYxIAA8qL3GFoHya:bw76raj) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L45)):*
> ```YAML
> 
>       - name: Build solution
>         run: msbuild build\SayMore.proj /t:Build /p:Configuration=$env:Configuration /p:useNUnit-x86=true /m
> ```

/p:useNunit-x86=true is definitely not needed here. This target doesn't run the NUnit3 command at all.
___
*[`.github/workflows/build-test-installer-release.yml` line 48 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-Ocqjnny17AqgMppi77z:-Ocqjnny17AqgMppi78-:b-w8cazi) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L48)):*
> ```YAML
> 
>       - name: Run tests
>         run: msbuild build\SayMore.proj /t:Test /p:Configuration=$env:Configuration /p:useNUnit-x86=true /p:excludedCategories=SkipOnTeamCity /m
> ```

I don't think /p:useNunit-x86=true is needed here. On TeamCity, I see no evidence that we are setting this. SayMore is currently a 32-bit app (though I intend to change that), and so I think the console runner automatically runs it as such. Forcing the --x86 flag here will probably just cause trouble when I switch it to be 64-bit.

tombogle (commented):
> @tombogle reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/243)* status: all files reviewed, 2 unresolved discussions (waiting on @megahirt)

megahirt:
@tombogle Thank you for this review.  Essentially, this PR is a minimal change to get GHA working, so that we no longer rely on TC to build, test, etc.  And then we can work on all the other things you mention in this review.

But I would prefer to merge this mostly as is and create tasks for each of the things you mentioned in this review:

- [ ] convert to SDK style csproj files so we can do `dotnet test`
- [ ] rename or reconsider "skipOnTeamCity" tests
- [ ] build project for 64-bit
- [ ] establish a robust version number system that is as close to automatic as possible
- [ ] ensure ConvertReleaseNotesToHtml is part of the release process

megahirt (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/saymore/243)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @tombogle)

___
*[`.github/workflows/build-test-installer-release.yml` line 20 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-Ocqnt-05KdOhpIUiynU:-OcrHJ4i5zMQqyXVo5De:btj0tg7) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L20)):*

Do we have control over the GH run number? In TeamCity, I reset the build number to the 0 in the production build every time I bump up the major or minor version number. It's actually annoying to have to remember to do this, so maybe I should bite the bullet and start using GitVersion. Also, it's kind of misleading to have 3.7 here. Currently, the major and minor version numbers actually come from the proj file. The TeamCity build configs Set BUILD_NUMBER as *.*.%build.counter%.{build.vcs.number.1}, so the major and minor "numbers" are literal asterisks, not numbers. Clearly, it would be better (if not switching to GitVersion) to change the proj file to not expect BUILD_NUMBER to just be %build.counter%.{build.vcs.number.1} (or maybe even just %build.counter%).
</blockquote></details>

GHA last time I checked doesn't have a run number / counter that we can "reset".  Which is unfortunately a missing feature that we used in TeamCity.  I would suggest that we fix up the version number in a separate PR, and that we see how automated we can make it.  I think GitVersion is overkill for small projects and we can do something simple, like take whatever the tag is named after the "v" and call that the version number.  We could do some env or file stamping stuff to get the version number in the right places for the build.
___
*[`.github/workflows/build-test-installer-release.yml` line 45 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-OcqjYxIAA8qL3GFoHy_:-OcrItNM0cq7fOIVmFaB:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L45)):*

/p:useNunit-x86=true is definitely not needed here. This target doesn't run the NUnit3 command at all.
</blockquote></details>

Done.
___
*[`.github/workflows/build-test-installer-release.yml` line 48 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-OcrAW5z4sLAjm6rzI3w:-OcrI5wUEsXEKRytroW4:b-l7ldqm) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L48)):*

Maybe it's time to convert the csproj file to SDK style so we can just use dotnet test.
</blockquote></details>

agreed.  Let's do this in a separate PR.
___
*[`.github/workflows/build-test-installer-release.yml` line 48 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-Ocqjnny17AqgMppi77z:-OcrItx47rsQyas501o9:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/.github/workflows/build-test-installer-release.yml#L48)):*

I don't think /p:useNunit-x86=true is needed here. On TeamCity, I see no evidence that we are setting this. SayMore is currently a 32-bit app (though I intend to change that), and so I think the console runner automatically runs it as such. Forcing the --x86 flag here will probably just cause trouble when I switch it to be 64-bit.
</blockquote></details>

Done.
___
*[`build/SayMore.proj` line 4 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-Ocr0m7eAWOnbk6b9W-j:-OcrIqLt1Uvmn2BHGJAH:b-896fix) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/build/SayMore.proj#L4)):*

I think we'll need to change this to:
</blockquote></details>

Done.
___
*[`build/SayMore.proj` line 100 at r2](https://reviewable.io/reviews/sillsdev/saymore/243#-Ocr4Uf9-aK4dPemL_k6:-OcrIBxQ6xlnvfw6aS6t:b-m32201) ([raw file](https://github.com/sillsdev/saymore/blob/44d722746d2e0fa898bc3fffb02ccf9517edab48/build/SayMore.proj#L100)):*

"SkipOnTeamCity" is a rather unfortunate category name since it is always (or very nearly always) used to indicate a test that probably can't be run in any CI scenario. We probably want to simplify this and remove the Condition (see Glyssen, et. al. for an example) and define and use ExtraExcludeCategories so that it contains SkipOnTeamCity for both flavors of CI build. Note: The TeamCity property (below), is specific to TC, so it will need to become TeamCity="$(TeamCity)", where TeamCity is defined as
</blockquote></details>

agreed - let's spend more time on this in a separate PR

